PR #32, opened 2021-10-08
Ensure correct results from responses order

yen-tt:
- for async methods that update the states (universal/vertical search and autocomplete), generate new id for every new request and check if the response is from the latest request before calling redux dispatch to update the state.
- added test for the case where response from older request comes after response from latest request

J=SLAP-1625
TEST=manual&auto
- see jest test pass with the added id tracking method, and fail without it
- hook to local headless react and see that searches still work and updated properly

### DIFF
--- a/tests/integration/query.ts
+++ b/tests/integration/query.ts
@@ -62,9 +62,16 @@ it('universal autocomplete sets search intents', async () => {
 
 describe('ensure correct results from latest request', () => {
   jest.useFakeTimers();
+  const queries = ['really long request', 'short', 'long request'];
+  const requestsTime = {
+    [queries[0]]: 20,
+    [queries[1]]: 5,
+    [queries[2]]: 12
+  };
+
   const mockAutoCompleteFn = jest.fn(
     async (request: VerticalAutocompleteRequest | UniversalAutocompleteRequest) => {
-      const waitTime = request.input?.length;
+      const waitTime = requestsTime[request.input];
       return new Promise(res => setTimeout(() => res({ results: [ {value: request.input} ]}), waitTime));
     }
   );
@@ -73,12 +80,12 @@ describe('ensure correct results from latest request', () => {
     verticalAutocomplete: mockAutoCompleteFn,
     universalAutocomplete: mockAutoCompleteFn,
     verticalSearch: jest.fn( async (request: VerticalSearchRequest) => {
-      const waitTime = request.query?.length;
+      const waitTime = requestsTime[request.query];
       return new Promise(res => setTimeout(() => res(
         { verticalResults: { results: [request.query] } }), waitTime));
     }),
     universalSearch: jest.fn( async (request: UniversalSearchRequest) => {
-      const waitTime = request.query?.length;
+      const waitTime = requestsTime[request.query];
       return new Promise(res => setTimeout(() => res(
         { verticalResults: [{ results: [request.query] }] }), waitTime));
     })
@@ -87,7 +94,6 @@ describe('ensure correct results from latest request', () => {
   const httpManager = new HttpManager();
   const statefulCore = new StatefulCore(mockedCore, stateManager, httpManager);
   statefulCore.setVerticalKey('someKey');
-  const queries = ['really long request', 'short', 'long request'];
   const updateResult = jest.fn();
 
   beforeEach(() => {
@@ -107,10 +113,10 @@ describe('ensure correct results from latest request', () => {
     statefulCore.setQuery(queries[2]);
     const thirdResponsePromise = statefulCore.executeVerticalAutoComplete();
 
-    jest.advanceTimersByTime(queries[1].length);
+    jest.advanceTimersByTime(requestsTime[queries[1]]);
     await secondResponsePromise;
     expect(statefulCore.state.vertical.autoComplete.results).toEqual([{ value: queries[1] }]);
-    jest.advanceTimersByTime(queries[2].length);
+    jest.advanceTimersByTime(requestsTime[queries[2]]);
     await thirdResponsePromise;
     jest.runAllTimers();
     await firstResponsePromise;
@@ -132,10 +138,10 @@ describe('ensure correct results from latest request', () => {
     statefulCore.setQuery(queries[2]);
     const thirdResponsePromise = statefulCore.executeUniversalAutoComplete();
 
-    jest.advanceTimersByTime(queries[1].length);
+    jest.advanceTimersByTime(requestsTime[queries[1]]);
     await secondResponsePromise;
     expect(statefulCore.state.universal.autoComplete.results).toEqual([{ value: queries[1] }]);
-    jest.advanceTimersByTime(queries[2].length);
+    jest.advanceTimersByTime(requestsTime[queries[2]]);
     await thirdResponsePromise;
     jest.runAllTimers();
     await firstResponsePromise;
@@ -157,10 +163,10 @@ describe('ensure correct results from latest request', () => {
     statefulCore.setQuery(queries[2]);
     const thirdResponsePromise = statefulCore.executeVerticalQuery();
 
-    jest.advanceTimersByTime(queries[1].length);
+    jest.advanceTimersByTime(requestsTime[queries[1]]);
     await secondResponsePromise;
     expect(statefulCore.state.vertical.results.verticalResults).toEqual({ results: [queries[1]] });
-    jest.advanceTimersByTime(queries[2].length);
+    jest.advanceTimersByTime(requestsTime[queries[2]]);
     await thirdResponsePromise;
     jest.runAllTimers();
     await firstResponsePromise;
@@ -182,10 +188,10 @@ describe('ensure correct results from latest request', () => {
     statefulCore.setQuery(queries[2]);
     const thirdResponsePromise = statefulCore.executeUniversalQuery();
 
-    jest.advanceTimersByTime(queries[1].length);
+    jest.advanceTimersByTime(requestsTime[queries[1]]);
     await secondResponsePromise;
     expect(statefulCore.state.universal.results.verticalResults).toEqual([{ results: [queries[1]] }]);
-    jest.advanceTimersByTime(queries[2].length);
+    jest.advanceTimersByTime(requestsTime[queries[2]]);
     await thirdResponsePromise;
     jest.runAllTimers();
     await firstResponsePromise;


src/stateful-core.ts
```diff
@@ -24,7 +24,23 @@ import { Unsubscribe } from '@reduxjs/toolkit';
 import { isLevenshteinMatch } from './utils/searchable-facets';
 
 export default class StatefulCore {
-  constructor(private core: AnswersCore, private stateManager: StateManager) {}
+  private latestRequestIds: Record<string, number>;
```
oshi97, line 27: it might make sense to pull this out into a new class

src/stateful-core.ts
```diff
@@ -24,7 +24,23 @@ import { Unsubscribe } from '@reduxjs/toolkit';
...
+  private getNewRequestId(requestName: string): number {
+    const newId = this.latestRequestIds[requestName]
+      ? this.latestRequestIds[requestName] + 1
+      : 1;
+    this.latestRequestIds[requestName] = newId;
+    return newId;
+  }
+
+  private getLatestRequestId(requestName: string): number {
+    return this.latestRequestIds[requestName];
```
oshi97, line 42: with the current way it's used it's impossible for this to happen, but if you call getLatestRequestId for a requestName that hasn't been set yet it'll return undefined, instead of a number. I was thinking it could be better if latestRequestIds starts off initialized with
```
{
  univeralQuery: 0,
  verticalQuery: 0
//etc
}
```
and we restrict requestName to one of the possible query names, that way we don't have to worry about it and also get nicer static typing. It also lets us simplify getNewRequestId as well (no need for defaulting)

src/stateful-core.ts
```diff
@@ -24,7 +24,23 @@ import { Unsubscribe } from '@reduxjs/toolkit';
...
 
 export default class StatefulCore {
-  constructor(private core: AnswersCore, private stateManager: StateManager) {}
+  private latestRequestIds: Record<string, number>;
+
+  constructor(private core: AnswersCore, private stateManager: StateManager) {
+    this.latestRequestIds = {};
+  }
+
+  private getNewRequestId(requestName: string): number {
```
oshi97, line 33: maybe `updateRequestId`? also okay with the current name though! think I'm splitting hairs here

yen-tt: yeah it does update request id internally, so your suggestion make sense. Updated!

src/request-manager.ts
```diff
@@ -0,0 +1,27 @@
+type RequestType = 'universalQuery'
```
oshi97, line 1: really nitting here, but maybe latestRequestIds can be a type of 
```ts
type RequestIds = {
  universalQuery: number,
  verticalQuery: number, //etc
}
```
and we can have
```ts
type RequestType = keyof RequestIds
```
It's really nitting but there's some chance of there being a typo in the initial keys of `this.latestRequestIds`. Another option is changing latestRequestIds to `Record<RequestType, number>`, though I don't think that would ensure that we are including ALL possible request types

tests/integration/query.ts
```diff
@@ -50,3 +58,130 @@ it('universal autocomplete sets search intents', async () => {
   expect(statefulCore.state.query.searchIntents).toEqual(['NEAR_ME']);
 });
 
+it('vertical autocomplete get latest request results', async () => {
+  const mockedCore: any = {
+    verticalAutocomplete: jest.fn(
+      async (_request: VerticalAutocompleteRequest) => {
```
oshi97, line 64: do we need to prefix request with an underscore? normally the underscore means the variable is unused

tests/integration/query.ts
```diff
@@ -50,3 +58,130 @@ it('universal autocomplete sets search intents', async () => {
...
+        const waitTime = _request.input?.length * 200;
```
oshi97, line 65: clever haha

tests/integration/query.ts
```diff
@@ -50,3 +58,130 @@ it('universal autocomplete sets search intents', async () => {
...
+  const queries = ['long request', 'short'];
+  statefulCore.setVerticalKey('someKey');
+
+  const dispatchEventSpy = jest.spyOn(stateManager, 'dispatchEvent');
+  statefulCore.setQuery(queries[0]);
+  statefulCore.executeVerticalAutoComplete();
+  statefulCore.setQuery(queries[1]);
+  statefulCore.executeVerticalAutoComplete();
+
+  await new Promise(res => setTimeout(res, 3000));
```
oshi97, line 81: instead of waiting an arbitrary time could we just wait for a `Promise.all` of the 2 autocompletes?

tests/integration/query.ts
oshi97: honestly we probably don't need to wait 200 seconds, I did some similar race conditions tests (that we ended up scraping) here -> https://github.com/yext/answers-headless/pull/25/commits/650eaedfeafcf2ed2f0613440a6524aedd09b612
and only waited 0, 1, and 2 ms.

src/request-manager.ts
```diff
@@ -0,0 +1,27 @@
...
+  | 'verticalQuery'
+  | 'universalAutoComplete'
+  | 'verticalAutoComplete';
+
+export default class RequestManager {
```
cea2aj, line 6: Can you add a jsdoc to describe the responsibility of the class?

src/request-manager.ts
```diff
@@ -0,0 +1,27 @@
...
+      universalAutoComplete: 0,
+      verticalAutoComplete: 0
+    };
+  }
+
+  updateRequestId(requestName: RequestType): number {
+    const newId = ++this.latestRequestIds[requestName];
+    this.latestRequestIds[requestName] = newId;
+    return newId;
+  }
```
cea2aj, line 22: We can simplify this function to just `return ++this.latestRequestIds[requestName];`

src/stateful-core.ts
```diff
@@ -131,6 +137,10 @@ export default class StatefulCore {
       referrerPageUrl
     });
 
+    const latestId = this.requestManager.getLatestRequestId('universalQuery');
+    if (currentId !== latestId) {
```
cea2aj, line 141: would this be a little clearer if it read: `if (thisRequestId !== latestRequestId) {`?

tests/integration/query.ts
```diff
@@ -50,3 +58,115 @@ it('universal autocomplete sets search intents', async () => {
   expect(statefulCore.state.query.searchIntents).toEqual(['NEAR_ME']);
 });
 
+
+describe('ensure correct results from latest request', () => {
+  const mockedCore: any = {
```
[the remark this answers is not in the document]
yen-tt: that wouldn't work since `new StatefulCore(mockedCore, stateManager)` expect it to be fully AnswersCore type

tests/integration/query.ts
```diff
@@ -50,3 +58,110 @@ it('universal autocomplete sets search intents', async () => {
...
+    jest.clearAllMocks();
+  });
+
+  it('vertical autocomplete get latest request results', async () => {
+    statefulCore.setQuery(queries[0]);
+    const firstResponsePromise = statefulCore.executeVerticalAutoComplete();
+    statefulCore.setQuery(queries[1]);
+    const secondResponsePromise = statefulCore.executeVerticalAutoComplete();
+
+    await Promise.all([firstResponsePromise, secondResponsePromise]);
```
cea2aj, line 104: Rather than waiting for both to resolve, could we use jest.advanceTimersByTime? For example, if we have one query take 10ms and another take 100ms, we could run `jest.advanceTimersByTime(10)` and check that shorter query has resolved and that the longer query has not yet resolved. And then we can run `jest.runAllTimers` so that the longer query is resolved. This way we know for sure the resolution order, and we also get the added benefit of not slowing down our tests by waiting for the timeouts

tests/integration/query.ts
```diff
@@ -50,3 +58,130 @@ it('universal autocomplete sets search intents', async () => {
...
+  statefulCore.executeUniversalAutoComplete();
+  statefulCore.setQuery(queries[1]);
+  statefulCore.executeUniversalAutoComplete();
+
+  await new Promise(res => setTimeout(res, 3000));
+
+  expect(statefulCore.state.query.query).toEqual(queries[1]);
+  expect(statefulCore.state.universal.autoComplete.results).toEqual([{ value: queries[1] }]);
+  expect(dispatchEventSpy).toBeCalledTimes(4);
+  expect(dispatchEventSpy).toHaveBeenNthCalledWith(1, 'query/set', queries[0]);
```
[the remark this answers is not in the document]
yen-tt: updated to use listener!

tests/integration/query.ts
yen-tt: oh that's cool. didn't know about that, I will try something out with it! In that case, it wouldn't matter much if we make the time gap to be small (5ms & 12ms) since this would just advance the time.

tests/integration/query.ts
```diff
@@ -50,3 +58,121 @@ it('universal autocomplete sets search intents', async () => {
   expect(statefulCore.state.query.searchIntents).toEqual(['NEAR_ME']);
 });
 
+
+describe('ensure correct results from latest request', () => {
+  jest.useFakeTimers();
+  const mockedCore: any = {
+    verticalAutocomplete: jest.fn(
+      async (request: VerticalAutocompleteRequest) => {
```
oshi97, line 66: I think it would be a little easier to read if we extract these jest.fn() instantiations to a single method, we could have it take in a typescript generic for the request type

yen-tt: I extracted the autocomplete jest.fn instantiations but for the universal/vertical search functions, I think we should keep it as is since their output is different.

oshi97: ah yeah I missed that!

tests/integration/query.ts
```diff
@@ -50,3 +59,139 @@ it('universal autocomplete sets search intents', async () => {
...
+    const firstResponsePromise = statefulCore.executeVerticalAutoComplete();
+    statefulCore.setQuery(queries[1]);
+    const secondResponsePromise = statefulCore.executeVerticalAutoComplete();
+    statefulCore.setQuery(queries[2]);
+    const thirdResponsePromise = statefulCore.executeVerticalAutoComplete();
+
+    jest.advanceTimersByTime(queries[1].length);
+    await secondResponsePromise;
+    expect(statefulCore.state.vertical.autoComplete.results).toEqual([{ value: queries[1] }]);
+    jest.advanceTimersByTime(queries[2].length);
```
oshi97, line 113: why do we advance from 5ms to 17ms, and then runAllTimers instead of just runAllTimers?

tests/integration/query.ts
```diff
@@ -50,3 +59,139 @@ it('universal autocomplete sets search intents', async () => {
...
+    jest.clearAllMocks();
+  });
+
+  it('vertical autocomplete get correct results based on up-to-date response', async () => {
+    statefulCore.addListener({
+      valueAccessor: state => state.vertical?.autoComplete?.results,
+      callback: updateResult
+    });
+
+    statefulCore.setQuery(queries[0]);
```
oshi97, line 103: nit but I feel like it's a little hard to read tests when they're written like this, since you have to mentally keep track of the what the value of the queries array is, and in this specific case also the length of each of the queries. Maybe we could have it so that the queries variable is a map of duration like "12 ms" to the query, or even just have the mock query be something like `"12"` and have the mock request parseInt that into the wait time

yen-tt: oh I updated so that the request time is a bit clearer compare to just extracting it from the length of the queries. But now that I read your comment again, you are suggesting to pull the shared queries (and time) into the individual tests? I think that would lead to some redundancy no? lmk if the update might make it easier, or if you still prefer setting the queries and time hardcoded in each test instead of outside scrope

PR #34, opened 2021-10-08
Add UniversalLimit support

cea2aj:
Add UniversalLimit support

Delete `setLimit` and create a `setVerticalLimit` and a `setUniversalLimit`.

`setVertialLimit` takes a string and applies to the current vertical. `setUniversalLimit` takes in a `UniversalLimit` object and can affect multiple verticals on universal search.

J=SLAP-1561
TEST=manual, auto

In the sample app, try calling `setVerticalLimit` and `setUniversalLimit` and see the searches properly limit the number of results. Add unit and integration tests.

### DIFF
--- a/src/stateful-core.ts
+++ b/src/stateful-core.ts
@@ -45,14 +45,12 @@ export default class StatefulCore {
     this.stateManager.dispatchEvent('vertical/setKey', key);
   }
 
-  setLimit(limit: number | UniversalLimit): void {
-    if (typeof limit === 'object') {
-      this.stateManager.dispatchEvent('universal/setLimit', limit);
-    } else if (typeof limit === 'number') {
-      this.stateManager.dispatchEvent('vertical/setLimit', limit);
-    } else {
-      console.error(`Unsupported limit type of ${typeof limit}`);
-    }
+  setVerticalLimit(limit: number): void {
+    this.stateManager.dispatchEvent('vertical/setLimit', limit);
+  }
+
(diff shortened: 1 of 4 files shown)


tests/integration/verticalsearch.ts
```diff
@@ -24,6 +24,12 @@ it('vertical searches update the search loading state', async () => {
   expect(statefulCore.state.vertical.searchLoading).toEqual(false);
 });
 
+it('statefulCore.setLimit sets the vertical limit when a string is passed to it', () => {
```
yen-tt, line 27: I think you mean when a number is passed to it, instead of a string

yen-tt (approved):
other than the small description change, lgtm!

src/stateful-core.ts
```diff
@@ -44,8 +45,14 @@ export default class StatefulCore {
     this.stateManager.dispatchEvent('vertical/setKey', key);
   }
 
-  setLimit(limit: number): void {
-    this.stateManager.dispatchEvent('vertical/setLimit', limit);
+  setLimit(limit: number | UniversalLimit): void {
```
oshi97, line 48: We might just want a separate action for setting the universal limit? The user will probably have to have some switching logic even if we have this internal if statement, since the parameter types are different, so I'm not sure if this makes people's lives easier.

cea2aj: I agree. I'll check with Tom on this

tmeyer2115 (approved):

PR #38, opened 2021-10-27
Support for multiple, isolated answers headless instances

oshi97:
This commit adds support for multiple answers headless instances, with a new `headlessId` in `provideAnswersHeadless`.

It does so by refactoring all slice ducks into "create" methods that take in an optional namespace.
This namespace is used to scope reducers scoped to a single `State` subtree. 
In order to do this, reducers are now managed by a HeadlessReducerManager class, which handles adding additional reducers for newly added instances, and recombining the reducer map.

In order for multiple StateManagers to share the same store, the store's instantiation was pulled out of ReduxStateManager. 
_(note: This has the nice side effect of fixing a buggy interaction we currently have with redux devtools in the answers-headless-react sample app, where two instances of AnswersHeadless would appear in the devtools drop down.)_

All headless instances have their SessionTrackingState linked together, as a requirement from product.
In the future, we can work on an interface to let people easily link together instances, or even just a flag
to turn off the default linking.

I also changed the dev tools display name to AnswersHeadless, instead of the default, which is whatever the current `window.title` is.

J=SLAP-1647
TEST=manual,auto

tested multiple instances of AnswersHeadless on the same page, and setting different parts of the query in multiple child AnswersHeadless instances
test the set-state action
test linking multiple AnswersHeadless instances together using addListener()


src/index.ts
```diff
@@ -12,13 +22,47 @@ type HeadlessConfig = AnswersConfig;
...
   const httpManager = new HttpManager();
 
-  return new AnswersHeadless(answersCore, stateManager, httpManager);
+  const headless = new AnswersHeadless(answersCore, stateManager, httpManager);
+  if (!firstHeadlessInstance) {
+    firstHeadlessInstance = headless;
+  } else {
+    // Two-way bind the current headless instances with the first one instantiated on the page.
+    // This way, all headless instances on a page will have their sessionTracking states linked.
+    // We have to be careful not to create an infinite loop here.
```
cea2aj, line 44: Are we avoiding the infinite loop because the listener is only fired when the value of the valueAccessor changes?

src/headless-reducer-manager.ts
```diff
@@ -0,0 +1,52 @@
...
+import createLocationSlice from './slices/location';
+import { ActionWithHeadlessId } from './store';
+
+/**
+ * Manages the current map of headless IDs to Reducers.
+ */
+export default class HeadlessReducerManager {
+  private headlessIdToReducer: Record<string, Reducer> = {};
+
+  addReducer(headlessId: string): void {
```
cea2aj, line 20: do you think `addAnswersReducer` might be a little clearer?

src/headless-reducer-manager.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.headlessIdToReducer[headlessId] = createAnswersReducer(headlessId + '/');
+  }
+
+  getCombinedReducer(): Reducer<ParentState, ActionWithHeadlessId> {
```
cea2aj, line 24: nit: could `getParentReducer` be a bit more consistent since we also have a `ParentState`?

src/index.ts
oshi97: Yup! We have to be careful about reference equality.
let's say we have a `first` and `second` headless. The flow looks like this
1. we call `first.setSessionId('blah')`
2. this triggers the listener to `first`'s sessionTracking, which calls `second.setState({...second.state, sessionTracking })`
4. because `second`'s sessionTracking has been updated, `second`'s sessionTracking listener calls `first.setState({...first.state, sessionTracking })`
5. We come back to `first`'s listener. Here, because we were careful about reference equality for sessionTracking, the listener sees that the reference has not changed, and the loop ends.

It's important to note that if we used the spread operator in the setState like 
```ts
first.setState({...first.state, sessionTracking: { ...sessionTracking } )
```
We would run into an infinite loop, or if the listener listened to the entire state instead of just `state.sessionTracking`

src/headless-reducer-manager.ts
oshi97: renamed to getParentReducer and addAnswersReducer!

src/index.ts
cea2aj: I see, we avoid the infinite loop since `sessionTracking` is referencing the same object in state

cea2aj (approved):

PR #42, opened 2021-11-01
Update how static filter is handled in headless

yen-tt:
This PR aim to solve two issues: 1) current setup can't support multiple static filters 2) StaticFilter have its own FilterState from answers-headless FilterState. This cause problems where applied filters have to use workaround (trigger click event) to update filter option for static filter. 

- update static in FilterState to handle support for multiple static filters using a map of `filterCollectionId` to a list of `SelectableFilter`s. When send to core, these chunks of filters will be transform to the nested structure and will be AND together.
- update static in FilterState to support `selected` option field
- update answers-headless and filter slice to include `setFilterOption`.
- add helper functions to convert filter structure a list of `SelectableFilter` to core format.
- add jest tests

will release version 0.1.0-beta.2 when merged

J=SLAP-1661
TEST=auto

see that jest tests passed

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -4,7 +4,6 @@ import {
   QuerySource,
   QuestionSubmissionRequest,
   Filter,
-  CombinedFilter,
   AutocompleteResponse,
   VerticalSearchResponse,
   UniversalSearchResponse,
@@ -26,6 +25,8 @@ import StateManager from './models/state-manager';
 import { Unsubscribe } from '@reduxjs/toolkit';
 import HttpManager from './http-manager';
 import answersUtilities from './answers-utilities';
+import { CombinedSelectableFilter, SelectableFilter } from './models/utils/selectablefilter';
+import { transformFiltersToCoreFormat } from './utils/transform-filters';
 
 export default class AnswersHeadless {
   readonly utilities: typeof answersUtilities;
@@ -66,7 +67,7 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
-  setFilter(filter: Filter | CombinedFilter | null): void {
+  setFilter(filter: Record<string, SelectableFilter | CombinedSelectableFilter> | null): void {
     this.stateManager.dispatchEvent('filters/setStatic', filter);
   }
 
@@ -194,7 +195,7 @@ export default class AnswersHeadless {
     const skipSpellCheck = !this.state.spellCheck.enabled;
     const sessionTrackingEnabled = this.state.sessionTracking.enabled;
     const sessionId = this.state.sessionTracking.sessionId;
-    const staticFilters = this.state.filters.static || undefined;
+    const staticFilters = transformFiltersToCoreFormat(this.state.filters.static) || undefined;
     const facets = this.state.filters?.facets;
     const limit = this.state.vertical.limit;
     const offset = this.state.vertical.offset;
@@ -296,6 +297,32 @@ export default class AnswersHeadless {
       facetOption
     };
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
+  }
+
+  addFilters(staticFiltersId: string, filters: Filter[]): void {
+    const payload = {
+      staticFiltersId,
+      filters
+    };
+    this.stateManager.dispatchEvent('filters/addFilters', payload);
+  }
+
+  selectFilterOption(filter: Filter, staticFiltersId?: string): void {
+    const payload = {
+      shouldSelect: true,
+      staticFiltersId,
+      filter
+    };
+    this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
+  }
+
+  unselectFilterOption(filter: Filter, staticFiltersId?: string): void {
+    const payload = {
+      shouldSelect: false,
+      staticFiltersId,
+      filter
+    };
+    this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
   }
 }
 
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -25,7 +25,7 @@ import StateManager from './models/state-manager';
 import { Unsubscribe } from '@reduxjs/toolkit';
 import HttpManager from './http-manager';
 import answersUtilities from './answers-utilities';
-import { CombinedSelectableFilter, SelectableFilter } from './models/utils/selectablefilter';
+import { SelectableFilter } from './models/utils/selectablefilter';
 import { transformFiltersToCoreFormat } from './utils/transform-filters';
 
 export default class AnswersHeadless {
@@ -67,7 +67,7 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
-  setFilter(filter: Record<string, SelectableFilter | CombinedSelectableFilter> | null): void {
+  setFilter(filter: Record<string, SelectableFilter[]> | null): void {
     this.stateManager.dispatchEvent('filters/setStatic', filter);
   }
 
@@ -299,7 +299,7 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
 
-  addFilters(filterSetId: string, filters: Filter[]): void {
+  addFilters(filterSetId: string, filters: SelectableFilter[]): void {
     const payload = {
       filterSetId,
       filters
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -296,11 +296,12 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
 
-  toggleFilterOption(filter: SelectableFilter, filterCollectionId: string): void {
+  toggleFilterOption(seletableFilter: SelectableFilter, filterCollectionId: string): void {
+    const { selected, ...filter } = seletableFilter;
     const payload = {
       filterCollectionId,
-      filter: filter.filter,
-      shouldSelect: filter.selected
+      filter: filter,
+      shouldSelect: selected
     };
     this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
   }
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -65,8 +65,8 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
-  setFilter(filter: Record<string, SelectableFilter[]> | null): void {
-    this.stateManager.dispatchEvent('filters/setStatic', filter);
+  setStaticFilters(filters: Record<string, SelectableFilter[]> | null): void {
+    this.stateManager.dispatchEvent('filters/setStatic', filters);
   }
 
   setFacets(facets: DisplayableFacet[]): void {
@@ -297,28 +297,11 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
 
-  addFilters(filterCollectionId: string, filters: SelectableFilter[]): void {
+  toggleFilterOption(filterCollectionId: string, filter: Filter, shouldSelect: boolean): void {
     const payload = {
       filterCollectionId,
-      filters
-    };
-    this.stateManager.dispatchEvent('filters/addFilters', payload);
-  }
-
-  selectFilterOption(filter: Filter, filterCollectionId?: string): void {
-    const payload = {
-      shouldSelect: true,
-      filterCollectionId,
-      filter
-    };
-    this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
-  }
-
-  unselectFilterOption(filter: Filter, filterCollectionId?: string): void {
-    const payload = {
-      shouldSelect: false,
-      filterCollectionId,
-      filter
+      filter,
+      shouldSelect
     };
     this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
   }
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -297,27 +297,27 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
 
-  addFilters(filterSetId: string, filters: SelectableFilter[]): void {
+  addFilters(filterCollectionId: string, filters: SelectableFilter[]): void {
     const payload = {
-      filterSetId,
+      filterCollectionId,
       filters
     };
     this.stateManager.dispatchEvent('filters/addFilters', payload);
   }
 
-  selectFilterOption(filter: Filter, filterSetId?: string): void {
+  selectFilterOption(filter: Filter, filterCollectionId?: string): void {
     const payload = {
       shouldSelect: true,
-      filterSetId,
+      filterCollectionId,
       filter
     };
     this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
   }
 
-  unselectFilterOption(filter: Filter, filterSetId?: string): void {
+  unselectFilterOption(filter: Filter, filterCollectionId?: string): void {
     const payload = {
       shouldSelect: false,
-      filterSetId,
+      filterCollectionId,
       filter
     };
     this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -296,14 +296,14 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
 
-  toggleFilterOption(seletableFilter: SelectableFilter, filterCollectionId: string): void {
+  setFilterOption(seletableFilter: SelectableFilter, filterCollectionId: string): void {
     const { selected, ...filter } = seletableFilter;
     const payload = {
       filterCollectionId,
       filter: filter,
       shouldSelect: selected
     };
-    this.stateManager.dispatchEvent('filters/toggleFilterOption', payload);
+    this.stateManager.dispatchEvent('filters/setFilterOption', payload);
   }
 }
 
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -3,7 +3,6 @@ import {
   QueryTrigger,
   QuerySource,
   QuestionSubmissionRequest,
-  Filter,
   AutocompleteResponse,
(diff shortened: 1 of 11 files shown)


src/answers-headless.ts
```diff
@@ -66,7 +67,7 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
-  setFilter(filter: Filter | CombinedFilter | null): void {
+  setFilter(filter: Record<string, SelectableFilter[]> | null): void {
```
cea2aj, line 70: Can we simplify this interface by having setFilter take in a single SelectableFilter? SelectableFilter would need to have it's own `filterSetId` or `filterGroupId`. If we want to allow an array, we could have a `setFilters` function

src/models/utils/selectablefilter.ts
```diff
@@ -0,0 +1,6 @@
+import { Filter } from '@yext/answers-core';
+
+export interface SelectableFilter {
```
cea2aj, line 3: Can this extend Filter rather than have a filter as a prop? That way it would be more similar to DisplayableFacet which extends Facet. Also, would the name `StaticFilter` make sense?

yen-tt: I think internally it would be cleaner/easier to keep filter as its own object in `SelectableFilter ` instead of spreading it out into multiple fields so when we don't have to remake a filter type when transforming it to core's nested filter structure and whatnots. But I also don't have a strong preference on this, so it can be change if you think it's still better!
As for the name, I think we should still keep it as `SelectableFilter` since it's not strictly specific to static, just that it holds a selected option (maybe can reuse for some other filters, likeFilterSearch?). `StaticFilter` name might be confusing since it's so similar to `StaticFilters` component

src/answers-headless.ts
yen-tt: from my understanding, `setFilter` purpose is to be able to change the entire `static` portion of `FilterState` previously so I kept the functionality as is. If user want to modify/add a new set of static filters (say another StaticFilter component), they would use `addFilters`. I could rename it to:
`setFilter(filter: Record<string, SelectableFilter[]> | null)` => `setFilters(...)`
`addFilters(filterSetId: string, filters: SelectableFilter[])` => `setFilter(...)`
so it's a bit more consistent. Does that help?

cea2aj: I'm not sure that we need both `setFilter` and `addFilters` since their functionality is so similar. If we were to have an `addFilters`, I would expect a `removeFilters` as well.  I'm wondering if it would make more sense to have just a single `setFilters` which would allow for setting everything at once.

src/answers-headless.ts
```diff
@@ -295,5 +296,31 @@ export default class AnswersHeadless {
...
+
+  addFilters(filterSetId: string, filters: SelectableFilter[]): void {
+    const payload = {
+      filterSetId,
+      filters
+    };
+    this.stateManager.dispatchEvent('filters/addFilters', payload);
+  }
+
+  selectFilterOption(filter: Filter, filterSetId?: string): void {
```
cea2aj, line 308: Product mentioned in the Answers Headless feedback sheet that they would like us to move away from selectFacetOption and unselectFacetOption to an interface which has selected as a boolean param. For example: `setFacetOptionSelected(selected, fieldId, option)`. Personally I'm not a fan of boolean flags in functions because the behavior is less clear when reading the code, however as a middle ground, perhaps we could have a single `setFilter`  rather than `selectFilterOption` and `unselectFilterOption` which takes in a SelectableFilter rather than a Filter

src/models/slices/filters.ts
```diff
@@ -1,7 +1,10 @@
-import { CombinedFilter, DisplayableFacet, Filter, SortBy } from '@yext/answers-core';
+import { DisplayableFacet, SortBy } from '@yext/answers-core';
+import { SelectableFilter } from '../utils/selectablefilter';
 
 export interface FiltersState {
-  static?: Filter | CombinedFilter | null;
+  static?: {
+   [filterSetId: string]: SelectableFilter[] | null;
```
cea2aj, line 6: To make sure I'm understanding this correctly, we are creating a concept of a filterSetId which groups together filters. If we were to have multiple static filter components, they would each use the filters associated with that filterSetId. Is that correct? Also, how will we decide which key to use for AppliedFilters which won't have a filterSetId?

src/answers-headless.ts
yen-tt: oh I see. In that case, I think we can just directly expose `toggleFilterOption` instead of wrapping it with `selectFilterOption` and `unselectFilterOption` to match what facet is doing. It would become like this:
`selectFilterOption(filter: SelectableFilter, filterSetId?: string): void`

src/models/slices/filters.ts
yen-tt: That's correct. For each static filter set, headless will store the corresponding id and a list of filters. As for AppliedFilters. It would run through all the sets to find the corresponding filter to unselect.

src/models/utils/selectablefilter.ts
cea2aj: Creating a simple public interface is more important than how simple the code is internally. As to what's the simpler public interface, I am leaning towards having `SelectableFilter` extend `Filter` because that way all the fields can be on one documentation page of what a filter is. However I think we should talk to product because it's ultimately up to them

cea2aj: and keeping `SelectableFilter` sgtm

src/answers-headless.ts
```diff
@@ -295,5 +296,31 @@ export default class AnswersHeadless {
     };
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
+
+  addFilters(filterSetId: string, filters: SelectableFilter[]): void {
```
cea2aj, line 300: I'm thinking that we should discuss the public interface of filters as a team and come to an agreement on it. Then we can present it to product, and once they have signed-off, then we come back to this PR to make updates

src/answers-headless.ts
yen-tt: That could work but I believe this would mean that all static filters would need get the current filter.static state and merge it with its new set of filters (which would cause it to subscribe to the entire static filterstate). So say we just have something like `setFilters(Record<string, SelectableFilter[]> 
 null)` which will directly set to be state.filter.static. Then a StaticFilter component would do something like:
 ```
 const staticFilters = useAnswersState(state => state.filter.static);
 useAnswersAction().setFilters({...staticFilters, someFilterId: someListOfFilters});
 ```

yen-tt: Update: keeping `setFilter` (now changed to `setStaticFilters`) as it's better to use this when applying filters through url instead of calling multiple toggle functions for each filter. Removed `addFilters`

src/models/utils/selectablefilter.ts
yen-tt: Update: Product is fine with filter as a field in SelectableFilter. Keeping `SelectableFilter` as the name

src/answers-headless.ts
yen-tt: update: removed both selectFilterOption and unselectFilterOption. Now we only have `toggleFilterOption`

src/answers-headless.ts
```diff
@@ -295,5 +296,14 @@ export default class AnswersHeadless {
     };
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
+
+  toggleFilterOption(filterCollectionId: string, filter: Filter, shouldSelect: boolean): void {
```
cea2aj, line 300: Is there a reason for having `Filter` and `shouldSelect` rather than allowing a `SelectableFilter` to be used here? It's part of the interface for SelectableFilter and if we used it here we would have more consistent interfaces. 

Side note, if we were to make this interface `toggleFilterOption(filter: SelectableFilter, filterCollectionId)`, then we could easily make `filterCollectionId` optional in the future

src/models/utils/selectablefilter.ts
cea2aj: I'm okay with leaving this interface as-is, however it's worth noting that even if this were updated so that it extends Filter, we'd wouldn't need to remake a filter type when transforming it to core's nested filter structure because SelectableFilter is still a Filter. We could pass in SelectableFilters to core as Filters and it should work without needing any additional transformations. The `selected` field wouldn't do anything which is a potential point of confusion, but typescript would be okay with it because of duck typing.

src/slices/filters.ts
```diff
@@ -10,8 +11,17 @@ interface FacetPayload {
   shouldSelect: boolean
 }
 
+interface FilterPayload {
+  filterCollectionId: string
```
cea2aj, line 15: filterCollectionId SGTM!

src/answers-headless.ts
yen-tt: ah yeah, forgot to SelectableFilter. As for the the order of params, I don't think we would have an unknown/optional `filterCollectionId` scenario right now, but I moved it to last argument anyway just in case we do, like you said

src/slices/filters.ts
```diff
@@ -47,6 +57,38 @@ const reducers = {
...
+  toggleFilterOption: (state: FiltersState, { payload }: PayloadAction<FilterPayload>) => {
+    if (!state.static) {
+      state.static = {};
+    }
+    const { filterCollectionId, filter: targetFilter, shouldSelect } = payload;
+    if (!state.static[filterCollectionId]) {
+      if (filterCollectionId && shouldSelect) {
+        state.static[filterCollectionId] = [];
+      } else {
+        console.warn(`invalid static filters id: ${filterCollectionId}`);
```
cea2aj, line 69: Could we simplify this a bit by moving the filterCollectionId check near the top of the function? That way we could have : `if (!filterCollectionId) { console.warn('invalid...') }` and then the logic here could remove the double if statement and instead have `if (!state.static[filterCollectionId] && shouldSelect) { ...`. We may not need the `shouldSelect` in the if statement either. If the filter isn't in the state and shouldSelect is false, I think it's okay to put that in the state so long as it's marked as not selected

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,71 @@
+import { CombinedFilter, Filter, FilterCombinator } from '@yext/answers-core';
+import { SelectableFilter } from '../models/utils/selectablefilter';
+
+function formatOrFilters(filters: Filter[]): Filter | CombinedFilter {
```
cea2aj, line 4: `formatOrFilters` makes me think that we are passing in 'Or' filters to this function and formatting them. How about a name such as 'combineFiltersWithOR'?

src/slices/filters.ts
yen-tt: if a user somehow unselect something that is not part of the state (never selected), that is an odd behavior no?

src/utils/transform-filters.ts
yen-tt: make sense. Update!

src/slices/filters.ts
yen-tt: ah yeah, will do

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
...
+      ? groups[element.filter.fieldId].push({ ...element.filter })
+      : groups[element.filter.fieldId] = [{ ...element.filter }];
+    return groups;
+  }, {});
+
+  const groupedFilterLabels = Object.keys(groupedFilters);
+  if (groupedFilterLabels.length === 1) {
+    return combineFiltersWithOR(groupedFilters[groupedFilterLabels[0]]);
+  }
+  return {
```
cea2aj, line 40: Assuming this behavior of ORing together filters with the same fileldId and ANDing together the filters grouped by filterId is the same as the SDK, this logic LGTM

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export function transformFiltersToCoreFormat(
+  filters: Record<string, SelectableFilter[]> | null | undefined
+): Filter | CombinedFilter | null {
+  if (!filters) {
+    return null;
+  }
+
+  const coreFormattedFilters: (Filter | CombinedFilter)[] = [];
+  Object.values(filters).forEach((filter: SelectableFilter[]) => {
```
cea2aj, line 58: nit: Can we use a plural `selectableFilters` rather than `filter` since it's an array?

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
...
+    const selectedFilters = filter.filter(filter => filter.selected);
```
cea2aj, line 59: nit: Can we use `selectableFilter` rather than `filter` since the type is a `SelectableFilter`?

src/utils/transform-filters.ts
yen-tt: I followed the logic we had for previous StaticFilters in React. But I just double check the SDK stuff (https://github.com/yext/answers-search-ui/blob/master/src/ui/components/filters/filteroptionscomponent.js#L719-L726) seems like the logic is the same so we should be good

tests/unit/slices/filters.ts
```diff
@@ -1,26 +1,225 @@
...
+    };
+
+    let actualState = reducer(initialState, toggleFilterOption(unselectFilterPayload));
+    const unselectExpectedState = _.cloneDeep(initialState);
+    unselectExpectedState.static.someId[1].selected = false;
+    expect(actualState).toEqual(unselectExpectedState);
+
+    actualState = reducer(initialState, toggleFilterOption(selectFilterPayload));
+    const selecteExpectedState = _.cloneDeep(initialState);
+    selecteExpectedState.static.someId[2].selected = true;
```
cea2aj, line 176: `selecte` rather than `selected`

yen-tt: oops, good catch

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
+import { CombinedFilter, Filter, FilterCombinator } from '@yext/answers-core';
+import { SelectableFilter } from '../models/utils/selectablefilter';
+
+/**
+ * Combine a given list of Filter using OR condition into a single CombinedFilter object
```
cea2aj, line 5: nit: I have a suggestion for wording: "Combine a list of Filters using the logical OR operator into a CombinedFilter"

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
...
+    return filters[0];
+  }
+  return {
+    combinator: FilterCombinator.OR,
+    filters: filters
+  };
+}
+
+/**
+ * Convert a list of SelectableFilter to nested filter structure use in core
```
cea2aj, line 18: another wording suggestion: 'Convert a list of SelectableFilters into the nested filter structure used by Core'

src/utils/transform-filters.ts
```diff
@@ -0,0 +1,74 @@
...
+    filters: Object.values(groupedFilters).map((filters: Filter[]) => combineFiltersWithOR(filters))
+  };
+}
+
+/**
+ * Convert a map of filtersId to filters in nested structure use in Answers Headless
+ * to a single nested filter stucture use in Answers Core
+ */
+export function transformFiltersToCoreFormat(
+  filerCollections: Record<string, SelectableFilter[]> | null | undefined
```
cea2aj, line 51: Missing the 't' in filter

src/models/utils/selectablefilter.ts
```diff
@@ -0,0 +1,6 @@
...
+  filter: Filter,
+  selected: boolean
```
oshi97, line 5: I know Connor already commented something similar, but I feel a little hesitant calling this a "SelectableFilter" if it doesn't extend "Filter". I think I would prefer just having it extend "Filter", since otherwise the state is a little more nested than it might need to be

oshi97: Also thanks to destructuring I don't think it should be tooo ugly to format the filters for core?

yen-tt: hmm ok, I'm fine either way. yeah it would be more consistent with the facets interface like connor said. Since the team prefer the extend approach, I can make the change!

src/slices/filters.ts
```diff
@@ -47,6 +57,34 @@ const reducers = {
...
+    });
+    if (foundFilter) {
+      foundFilter.selected = shouldSelect;
+    } else if (shouldSelect) {
+      const selectedFilter = { filter: targetFilter, selected: shouldSelect };
+      filterCollection
+        ? filterCollection.push(selectedFilter)
+        : state.static[filterCollectionId] = [selectedFilter];
+    } else {
+      console.warn('Could not unselect a non-existing filter option in state'
```
oshi97, line 84: It might be helpful to include the filterCollectionId here?

src/slices/filters.ts
```diff
@@ -47,6 +57,34 @@ const reducers = {
...
+    }
+    const { filterCollectionId, filter: targetFilter, shouldSelect } = payload;
+    if (!filterCollectionId) {
+      console.warn(`invalid static filters id: ${filterCollectionId}`);
+      return;
+    }
+    const filterCollection = state.static[filterCollectionId];
+    const foundFilter = filterCollection?.find(storedSelectableFilter => {
+      const storedFilter = storedSelectableFilter.filter;
+      return storedFilter.fieldId === targetFilter.fieldId
```
oshi97, line 72: in the future maybe we can export a filter equality method or something as part of AnswersUtilities (not right now just thinking it might be useful in the future for people)

oshi97 (commented):
sheeesh that's some test coverage gurl

src/slices/filters.ts
yen-tt: good point, updated!

src/models/utils/selectablefilter.ts
yen-tt: updated

src/slices/filters.ts
yen-tt: sounds good

oshi97 (approved):

src/answers-headless.ts
```diff
@@ -295,5 +295,15 @@ export default class AnswersHeadless {
     };
     this.stateManager.dispatchEvent('filters/toggleFacetOption', payload);
   }
+
+  toggleFilterOption(seletableFilter: SelectableFilter, filterCollectionId: string): void {
```
cea2aj, line 299: nit: we may want to call this `setFilterOption`, because to me a toggle should already know what the current state is, and simply flip it. This seems to be more of a set since we are specifying whether or not the filter should be selected

cea2aj:
Overall looks great

src/answers-headless.ts
yen-tt: that make sense. updated

cea2aj (approved):

PR #43, opened 2021-11-08
Top level direct answer slice

cea2aj:
Create a slice for direct answers at the top level of the state

I originally tried using the direct answer models as the state of the slice, however I was running into problems with redux when attempting it. I was able to get it working by putting the direct answer under a prop which I called 'result'. I believe that putting the direct answer under a prop is the best way of maintaining the typescript discriminating union. I think It reads well, and it gives us flexibility in case we need to add anything to the direct answer state which is specific to headless.

J=SLAP-1664
TEST=auto

Add unit test which confirms the direct answer is set properly in this new slice

### DIFF
--- a/src/models/slices/directanswer.ts
+++ b/src/models/slices/directanswer.ts
@@ -1,0 +1,5 @@
+import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
+
+export interface DirectAnswerState {
+  result?: FeaturedSnippetDirectAnswer | FieldValueDirectAnswer;
+}
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -1,8 +1,9 @@
-The following NPM package may be included in this product:
+The following NPM packages may be included in this product:
 
  - @babel/runtime-corejs3@7.15.4
+ - @babel/runtime@7.14.6
 
-This package contains the following license and notice below:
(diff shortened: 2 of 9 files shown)


tmeyer2115 (approved):

src/models/slices/directanswer.ts
```diff
@@ -0,0 +1,5 @@
+import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
+
+export interface DirectAnswerState {
+  result?: FeaturedSnippetDirectAnswer | FieldValueDirectAnswer;
```
yen-tt, line 4: I think result can be `undefined` as well for the types

cea2aj: The question mark indicates that result may be undefined. Are you suggesting to make this  `FeaturedSnippetDirectAnswer | FieldValueDirectAnswer | undefined`?

yen-tt: hm ok, I wasn't sure if the question mark would cover the case when we directly set it as undefined `state.result = action.payload;` compare to just it being undefined because it was never set. If that's the same thing, then all good!

yen-tt (approved):

src/models/slices/directanswer.ts
cea2aj: yep, they are the same

PR #44, opened 2021-11-08
Add support for FilterSearch

yen-tt:
Add support for filterSearch data in state

- added filterSearch following interface `FilterSearchStates` to top level `State`, which contains a mapping of filterSearchId to `FilterSearchState` with query and results field. This is to support multiple filter search components
- add `setFilterSearchQuery` as part of headless public interface. Update `executeFilterSearch` to store results in state, and maintain correct response order
- added reducers
- added more typing, and tests

the flow for React component would be: use `setFilterSearchQuery` to update the input value of each filterSearch component, `executeFilterSearch` on submit, and call `toggleFilterOption` to add the selected filter for a specific filterSearch component from the list of filterSearch results.

J=SLAP-1685
TEST=auto

see that jest test passed

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -283,7 +283,6 @@ export default class AnswersHeadless {
   ): Promise<FilterSearchResponse | undefined> {
     const thisRequestId = this.httpManager.updateRequestId('filterSearch');
     const verticalKey = this.state.vertical.key;
-    console.log(verticalKey);
     if (!verticalKey) {
       console.error('no verticalKey supplied for filter search');
       return;
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -14,7 +14,6 @@ import {
   Context,
   LatLong,
   SearchParameterField,
-  FilterSearchResponse,
   UniversalLimit
 } from '@yext/answers-core';
 
@@ -280,7 +279,7 @@ export default class AnswersHeadless {
     filterSearchId: string,
     sectioned: boolean,
     fields: SearchParameterField[]
-  ): Promise<FilterSearchResponse | undefined> {
+  ): Promise<void> {
     const thisRequestId = this.httpManager.updateRequestId('filterSearch');
     const verticalKey = this.state.vertical.key;
     if (!verticalKey) {
@@ -308,7 +307,7 @@ export default class AnswersHeadless {
 
     const latestResponseId = this.httpManager.getLatestResponseId('filterSearch');
     if (thisRequestId < latestResponseId) {
-      return results;
+      return;
     }
     const payload = {
       filterSearchId,
@@ -316,7 +315,6 @@ export default class AnswersHeadless {
     };
     this.httpManager.setResponseId('filterSearch', thisRequestId);
     this.stateManager.dispatchEvent('filterSearch/setResults', payload);
-    return results;
   }
 
   selectFacetOption(fieldId: string, facetOption: FacetOption): void {
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -60,6 +60,14 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('universal/setLimit', limit);
   }
 
+  setFilterSearchQuery(query: string, filterSearchId: string): void {
+    const payload = {
+      query,
+      filterSearchId
+    };
+    this.stateManager.dispatchEvent('filterSearch/setQuery', payload);
+  }
+
   setOffset(offset: number): void {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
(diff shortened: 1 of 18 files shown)


src/answers-headless.ts
```diff
@@ -268,14 +276,42 @@ export default class AnswersHeadless {
...
-    return this.core.filterSearch({
-      input: this.state.query.query || '',
+  async executeFilterSearch(
+    filterSearchId: string,
+    sectioned: boolean,
+    fields: SearchParameterField[]
+  ): Promise<FilterSearchResponse | undefined> {
+    const thisRequestId = this.httpManager.updateRequestId('filterSearch');
+    const verticalKey = this.state.vertical.key;
+    console.log(verticalKey);
```
cea2aj, line 286: console log left in here

yen-tt: oh my bad

src/models/slices/filtersearch.ts
```diff
@@ -0,0 +1,10 @@
+import { FilterSearchResponse } from '@yext/answers-core';
+
+interface FilterSearchState {
+  query: string,
+  results?: FilterSearchResponse
```
cea2aj, line 5: Since we are removing the response models from answers-headless, can we instead spread the relevant properties into the FilterSearchState? We'll want sectioned, sections, results, and inputIntents (in headless we've been calling them searchIntents, so I think we should go with that for consistency). Also, the queryId should update state.query.queryId and the UUID should update state.meta.uuid (I'm still working on that change for regular searches and for autocomplete)

src/models/slices/filtersearch.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export interface FilterSearchStates {
+  [filterSearchId: string]: FilterSearchState
```
cea2aj, line 9: I'm thinking we might want to call this filterSearchCollectionId so that we're consistent with filterCollectionId

yen-tt: I was thinking the value here is not really a collection since it's a state of query + result, unlike the mapping in staticFilter where it's a collection/list of selectable filters, so maybe that naming wouldn't fit so well.

src/models/slices/filtersearch.ts
yen-tt: > Since we are removing the response models from answers-headless, can we instead spread the relevant properties into the FilterSearchState? We'll want sectioned, sections, results, and inputIntents (in headless we've been calling them searchIntents, so I think we should go with that for consistency). 
oh I see, will do

> Also, the queryId should update state.query.queryId and the UUID should update state.meta.uuid (I'm still working on that change for regular searches and for autocomplete)
Not sure if I follow this right, but the query from filterSearch is different from the other queries store in QueryState so would storing queryId in there make sense? similar thing with uuid.

src/answers-headless.ts
```diff
@@ -268,14 +276,41 @@ export default class AnswersHeadless {
...
+    fields: SearchParameterField[]
+  ): Promise<FilterSearchResponse | undefined> {
+    const thisRequestId = this.httpManager.updateRequestId('filterSearch');
+    const verticalKey = this.state.vertical.key;
+    if (!verticalKey) {
+      console.error('no verticalKey supplied for filter search');
+      return;
+    }
+    if (!filterSearchId || !this.state.filterSearch[filterSearchId]) {
+      console.error(`invalid filterSearch id: "${filterSearchId}""`);
```
cea2aj, line 291: Can this error explain how the id is invalid?

src/models/slices/filtersearch.ts
cea2aj: in that case I think  `filterSearchId` works

src/models/slices/filtersearch.ts
cea2aj: That's a good point that it doesn't really make sense to put it in the top level query since this has it's own different query. I'd keep queryId and uuid in the FilterSearchState for now then. I'm going to discuss this with Tom

src/answers-headless.ts
```diff
@@ -268,14 +276,41 @@ export default class AnswersHeadless {
...
+    if (thisRequestId < latestResponseId) {
+      return results;
+    }
+    const payload = {
+      filterSearchId,
+      results
+    };
+    this.httpManager.setResponseId('filterSearch', thisRequestId);
+    this.stateManager.dispatchEvent('filterSearch/setResults', payload);
+    return results;
```
cea2aj, line 313: Since we are removing the Response models, for now I think it makes sense to not return anything here. Again, I'll discuss with Tom when I get the chance

src/answers-headless.ts
yen-tt: updated, should be clearer now

src/answers-headless.ts
yen-tt: sounds good

src/answers-headless.ts
```diff
@@ -60,6 +59,14 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('universal/setLimit', limit);
   }
 
+  setFilterSearchQuery(query: string, filterSearchId: string): void {
```
tmeyer2115, line 62: I'm wondering if there's value in storing the filterSearchQuery in state actually. The normal search query, I think makes sense to store. There are a number of consumers of it and applications for it. I get the feeling that anyone setting a filterSearch query will want to fire a filterSearch almost immediately after. 

What would you think about removing this from the state for now and have it be passed into `executeFilterSearch`?

src/answers-headless.ts
```diff
@@ -270,14 +277,46 @@ export default class AnswersHeadless {
...
   }
 
-  executeFilterSearch(sectioned: boolean, fields: SearchParameterField[]): Promise<FilterSearchResponse> {
-    return this.core.filterSearch({
-      input: this.state.query.query || '',
+  async executeFilterSearch(
+    filterSearchId: string,
+    sectioned: boolean,
+    fields: SearchParameterField[]
+  ): Promise<void> {
```
tmeyer2115, line 284: Can we continue to return a `Promise<FilterSearchResponse>`?

src/models/slices/filtersearch.ts
```diff
@@ -0,0 +1,18 @@
+import { AutocompleteResult, SearchIntent } from '@yext/answers-core';
+
+export interface FilterSearchState {
+  query: string,
+  sectioned?: boolean,
+  sections?: {
+    label: string,
+    results: AutocompleteResult[]
+  }[],
+  results?: AutocompleteResult[],
```
tmeyer2115, line 10: I think it's confusing that people can access `results` at two different levels of abstraction. Could we always have `sections`? In the case where the back-end returns `sectioned` false, we have a single element in the Array. Open to other ideas as well, but I think our state design should aim to minimize the number of conditionals in the consumer's code.

src/answers-headless.ts
cea2aj: I suggested we remove it since we were removing the response models, however after discussing this further we decided to leave this functionality in

src/models/slices/filtersearch.ts
yen-tt: I did some more testing and looking at the [endpoint documentation](https://hitchhikers.yext.com/docs/liveapis/answersapi/verticalsearch/#operation/filtersearch). It doesn't seem like `results` (the one that's not nested inside `sections`) is used by the backend for filterSearch. If `sectioned` is false. Everything will be returned in the same section. We should update answers-core docs to reflect this behavior. I will remove the top level `results` from this pr

src/answers-headless.ts
yen-tt: all good, I will add it back in!

yen-tt:
From discussion with team and product, it's decided that filterSearch and autoComplete should be state-less operations.

PR #45, opened 2021-11-09
Remove response model from vertical state

cea2aj:
Update the vertical slice state so that the VerticalResults are spread into the state

This PR removes the vertical search response model which removes duplication in the state, and reduces the layers in the state

J=SLAP-1664
TEST=auto

Add unit tests which cover these new parts of state

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -14,7 +14,8 @@ import {
   LatLong,
   SearchParameterField,
   FilterSearchResponse,
-  UniversalLimit
+  UniversalLimit,
+  VerticalSearchResponse
 } from '@yext/answers-core';
 
 import StateListener from './models/state-listener';
@@ -180,7 +181,7 @@ export default class AnswersHeadless {
     return results;
   }
 
-  async executeVerticalQuery(): Promise<void> {
+  async executeVerticalQuery(): Promise<VerticalSearchResponse> {
     const thisRequestId = this.httpManager.updateRequestId('verticalQuery');
     const verticalKey = this.state.vertical.verticalKey;
     if (!verticalKey) {
@@ -249,6 +250,7 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('vertical/setResults', response.verticalResults?.results);
     this.stateManager.dispatchEvent('vertical/setResultsCount', response.verticalResults?.resultsCount);
     this.stateManager.dispatchEvent('vertical/setSource', response.verticalResults?.source);
+    return response;
   }
 
   async executeVerticalAutoComplete(): Promise<AutocompleteResponse | undefined> {
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -274,7 +274,7 @@ export default class AnswersHeadless {
     sectioned: boolean,
     fields: SearchParameterField[]
   ): Promise<FilterSearchResponse | undefined> {
-    const verticalKey = this.state.vertical.key;
+    const verticalKey = this.state.vertical.verticalKey;
     if (!verticalKey) {
       console.error('no verticalKey supplied for filter search');
       return;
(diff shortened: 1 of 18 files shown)


src/answers-headless.ts
```diff
@@ -226,30 +226,37 @@ export default class AnswersHeadless {
...
+    this.stateManager.dispatchEvent('filters/setFacets', response.facets);
+    this.stateManager.dispatchEvent('spellCheck/setResult', response.spellCheck);
+    this.stateManager.dispatchEvent('location/setLocationBias', response.locationBias);
+    this.stateManager.dispatchEvent('query/setSearchIntents', response.searchIntents || []);
+    this.stateManager.dispatchEvent('location/setLocationBias', response.locationBias);
+    this.stateManager.dispatchEvent('directAnswer/setResult', response.directAnswer);
     this.stateManager.dispatchEvent('vertical/setSearchLoading', false);
-    this.stateManager.dispatchEvent('directAnswer/setResult', results.directAnswer);
-    return results;
+    this.stateManager.dispatchEvent('vertical/setAllResultsForVertical', response.allResultsForVertical);
```
oshi97, line 244: what do you think about trying the single reducer you mentioned, for spreading the response into vertical state? feels like it might be a little nicer that way, also maybe slightly more performant since each dispatch loops through all the listeners (this is probably negligible though)

cea2aj: Sure, I can give it a shot

oshi97 (approved):

src/models/slices/vertical.ts
```diff
@@ -1,11 +1,34 @@
...
+  AutocompleteResponse,
+  DisplayableFacet,
+  Result,
+  SearchIntent,
+  SortBy,
+  Source,
+  VerticalResults
+} from '@yext/answers-core';
+
+export interface AllResultsForVertical {
```
tmeyer2115, line 12: Just refreshing my memory, this is what we decided would be best with Rose, correct? Later on, we may make a setting or something that automatically hydrates the top-level facets and search intents from this when there are no results.

src/models/slices/vertical.ts
```diff
@@ -1,11 +1,34 @@
...
+  results: Result[],
+  resultsCount: number,
+  searchIntents: SearchIntent[]
+}
+
 export interface VerticalSearchState {
-  alternativeVerticals?: VerticalResults[]
-  key?: string,
+  allResultsForVertical?: AllResultsForVertical,
+  alternativeVerticals?: VerticalResults[],
```
tmeyer2115, line 21: I'm wondering if we should group these two under an optional attribute called `noResults?` or something. The current state design doesn't make it clear when we can or can't expect this data. I suppose we could also solve that problem with documentation.

src/models/slices/vertical.ts
```diff
@@ -1,11 +1,34 @@
...
+  appliedQueryFilters?: AppliedQueryFilter[],
   autoComplete?: AutocompleteResponse,
-  results?: VerticalSearchResponse,
   displayName?: string
   limit?: number,
   offset?: number,
-  searchLoading?: boolean
+  queryDurationMillis?: number,
+  results?: Result[],
+  resultsCount?: number,
```
tmeyer2115, line 29: Minor nit: can we get rid of this? If they have the array of Results, they can deduce the count.

cea2aj: I believe we need to have it separately because there could be a resultsCount of 500 but the backend only returns 20 of them

src/models/slices/vertical.ts
cea2aj: Ya, this what we discussed we'd do for now. The only difference is that I changed the model name from `AllResults` to `AllResultsForVertical` since that's what the API uses

src/models/slices/vertical.ts
cea2aj: Currently, client code trying to use alternativeVerticals or allResultsForVertical would look something like: `if (state.vertical.resultsCount === 0 && state.vertical.alternativeVerticals)`, which reads well to me. However I don't have a strong preference here. I can ping Rose

cea2aj: Rose liked the noResults prop so i'll update!

src/models/slices/vertical.ts
tmeyer2115: Ahh, that's right! Nvmd

tmeyer2115 (approved):

PR #47, opened 2021-11-09
Remove ID from staticFilter state

yen-tt:
update StaticFilter to not use ids to manage multiple filters. Instead, all multiple filters are grouped into one array. The merging process for filters (AND/OR) will remain in headless, as discussed with product.

- remove mapping from FilterState, replace by list of SelectableFilter
- update public interface to remove filterCollectionId
- update transformFiltersToCoreFormat to be from list of SelectableFilter, instead of a mapping of id to filters, to the nested structure for core
- update jest tests

J=SLAP-1688
TEST=manual&auto

- see that updated jest tests passed
- hook to sample app, see that multiple static filters still work

### DIFF
--- a/src/slices/filters.ts
+++ b/src/slices/filters.ts
@@ -2,6 +2,7 @@ import { createSlice, PayloadAction, Slice } from '@reduxjs/toolkit';
 import { Filter, FacetOption, DisplayableFacet, SortBy } from '@yext/answers-core';
 import { FiltersState } from '../models/slices/filters';
 import { SelectableFilter } from '../models/utils/selectablefilter';
+import { areFiltersEqual } from '../utils/filter-utils';
 
 const initialState: FiltersState = {};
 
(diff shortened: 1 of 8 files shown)


src/slices/filters.ts
```diff
@@ -59,15 +58,10 @@ const reducers = {
...
-    if (!filterCollectionId) {
-      console.warn(`invalid static filters id: ${filterCollectionId}`);
-      return;
-    }
-    const filterCollection = state.static[filterCollectionId];
-    const foundFilter = filterCollection?.find(storedSelectableFilter => {
+    const { filter: targetFilter, shouldSelect } = payload;
+    const foundFilter = state.static.find(storedSelectableFilter => {
       const { selected:_, ...storedFilter } = storedSelectableFilter;
       return storedFilter.fieldId === targetFilter.fieldId
```
tmeyer2115, line 66: Could we make this conditional into a method? Something like `areFiltersEqual`?

src/slices/filters.ts
```diff
@@ -61,14 +62,12 @@ const reducers = {
       state.static = [];
     }
     const { filter: targetFilter, shouldSelect } = payload;
-    const foundFilter = state.static.find(storedSelectableFilter => {
+    const matchedFilter = state.static.find(storedSelectableFilter => {
```
tmeyer2115, line 65: Minor nit, could we say `matchingFilter`?

yen-tt: yup

tmeyer2115 (approved):

PR #49, opened 2021-11-10
Miscellaneous state improvements

cea2aj:
A collection of small state design improvements

These changes include:
- Add UUID to meta slice
- Move searchLoading to the top level of state under the status slice
- Rename state.query.query to state.query.input
- Rename query.latest to query.mostRecentSearch
- Rename vertical.key to vertical.verticalKey

J=SLAP-1664
TEST=auto

Update and add relevant test coverage


src/answers-headless.ts
```diff
@@ -36,8 +36,8 @@ export default class AnswersHeadless {
     private httpManager: HttpManager,
   ) {}
 
-  setQuery(query: string): void {
-    this.stateManager.dispatchEvent('query/set', query);
+  setQueryInput(input: string): void {
```
cea2aj, line 39: I updated this to be consistent with our convention of naming the setters 'setQuery + prop', however I'm not the biggest fan of this naming so I'm considering keeping `setQuery`. I'm not sure if that would make it unclear where the corresponding data is in the state. Let me know your thoughts

tmeyer2115: I like the idea of keeping it `setQuery`. The term `setQueryInput` has a front-end connotation to me. It sounds very much like I'm setting the value of the `<input>` tag for a search bar.

src/models/slices/status.ts
```diff
@@ -0,0 +1,3 @@
+export interface StatusState {
```
tmeyer2115, line 1: I think we should try and come up with a more descriptive name. As a consumer of the library, I would wonder "status of what"?

cea2aj: maybe `SearchStatusState`?

tmeyer2115: Sure, that sounds good

cea2aj: updated!

src/answers-headless.ts
cea2aj: reverted to `setQuery`

yen-tt (approved):

PR #50, opened 2021-11-10
update Vertical/Universal state (autocomplete and noResults scenario)

yen-tt:
- remove autocomplete from universal and vertical state, and the corresponding jest tests
- when merge with main, made updates to how verticals search response is handle to resolve typescript issues
  - allResultsForVertical and alternativeVerticals will both be present if there are no actual results. Code in main allowed for case when only one is present but interface expects both.

will publish new version once this is merged

J=SLAP-1690
TEST=auto

jest test


tests/integration/verticalsearch.ts
```diff
@@ -22,27 +22,49 @@ const initialState: State = {
   location: {}
 };
 
-it('vertical searches set allResultsForVertical', async () => {
+const alternativeVerticals = [{
```
cea2aj, line 25: nit: I usually add the corresponding typescript type to these variables just so that we know that the data is shaped correctly (same for allResultsForVertical)

src/answers-headless.ts
```diff
@@ -165,21 +165,11 @@ export default class AnswersHeadless {
     return response;
   }
 
-  async executeUniversalAutoComplete(): Promise<AutocompleteResponse | undefined> {
-    const thisRequestId = this.httpManager.updateRequestId('universalAutoComplete');
+  async executeUniversalAutoComplete(): Promise<AutocompleteResponse> {
```
cea2aj, line 168: While we're here, can we rename this to `executeUniversalAutocomplete`? Autocomplete is one word, so the 'C' should not be capitalized. In Core we don't capitalize the 'C'

src/answers-headless.ts
```diff
@@ -165,21 +165,11 @@ export default class AnswersHeadless {
...
     const query = this.state.query.input || '';
-    const results = await this.core.universalAutocomplete({
+    return this.core.universalAutocomplete({
```
cea2aj, line 170: I'm wondering if we want to return a different response model than the one from Core, however I think this works

yen-tt: I think it makes sense to return as is with the updated filterSearchResponse interface in core, which now should only include `sections` and some ids (queryId, businessId, and uuid).

cea2aj (approved):

PR #54, opened 2021-11-23
Support autocompleteSessionId

yen-tt:
Update headless state and network calls to support autocompleteSessionId

- Update `QueryState` to include SearchAggregation, which includes 'enabled' and 'id' field. This is similar to SessionTrackingState, and would give user better context in what it is use for compare to the backend name of 'autocompleteSessionId'.
- headless interface exposes two new functions, setSearchAggregationEnabled and setSearchAggregationId. Once searchAggregation is set to true, headless would internally setup new uuid and include it in every group of universal and vertical autocomplete/search calls. A new UUID will automatically be generated after a query is executed. However, in the case of id needing to update manually (i.e. when user type stuff in search bar but click elsewhere without executing search and the searchbar loses focus), then setSearchAggregationId can be used.
- added new tests

NOTE: product wants to hold off on publishing this in core latest or beta version until backend work is done.

J=SLAP-1723
TEST=auto

use updated core in headless, see that jest tests passed.

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -49,10 +49,6 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('query/setSource', source);
   }
 
-  setSearchAggregationEnabled(enabled: boolean): void {
-    this.stateManager.dispatchEvent('query/setSearchAggregationEnabled', enabled);
-  }
-
   setSearchAggregationId(uuid: string): void {
     this.stateManager.dispatchEvent('query/setSearchAggregationId', uuid);
   }
@@ -143,6 +139,7 @@ export default class AnswersHeadless {
     const sessionId = this.state.sessionTracking.sessionId;
     const { referrerPageUrl, context } = this.state.meta;
     const { userLocation } = this.state.location;
+    const autocompleteSessionId = this.state.query.searchAggregationId;
 
     const response = await this.core.universalSearch({
       query: input || '',
@@ -155,8 +152,7 @@ export default class AnswersHeadless {
       location: userLocation,
       context,
       referrerPageUrl,
-      ...(this.state.query.searchAggregation?.enabled 
-          && { autocompleteSessionId: this.state.query.searchAggregation.id })
+      autocompleteSessionId
     });
 
     const latestResponseId = this.httpManager.getLatestResponseId('universalQuery');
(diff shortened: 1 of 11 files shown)


src/answers-headless.ts
```diff
@@ -162,13 +173,18 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('searchStatus/setIsLoading', false);
     this.stateManager.dispatchEvent('meta/setUUID', response.uuid);
     this.stateManager.dispatchEvent('directAnswer/setResult', response.directAnswer);
+    if (this.state.query.searchAggregation?.enabled) {
+      this.stateManager.dispatchEvent('query/setSearchAggregationId', uuidv4());
```
oshi97, line 177: could you expand a little on how searches will be grouped together? It looks like here we would generate a new autocompleteSessionId for every single universal search made?

tests/integration/query.ts
```diff
@@ -118,3 +118,83 @@ describe('ensure correct results from latest request', () => {
     expect(updateResult.mock.calls).toHaveLength(2);
   });
 });
+
+
+let uuidString = 'some-uuid-value';
+const mockUuid = jest.fn(() => uuidString);
+jest.mock('uuid', () => ({
```
oshi97, line 125: would it be possible to do this with a `__mocks__/uuid` file, kind of like `__mocks__/@yext/answers-core`?

yen-tt: update: no longer required. it's removed

src/answers-headless.ts
yen-tt: After discussing with product, the usage of autocompleteSessionId is different from the functionality added in this pr. autocompleteSessionId should be used in vertical/universal search only when it's executed for the purpose of visual autocomplete (the search term aggregation and this id doesn't apply for normal autocomplete). My original thought was to update uuid after every final search is triggered, and the visual autocomplete would have another function to execute autocomplete + search endpoints. Updated to always include autocompleteSessionId, and user can set that if they want a uuid to be included in the request

src/answers-headless.ts
```diff
@@ -48,6 +48,10 @@ export default class AnswersHeadless {
     this.stateManager.dispatchEvent('query/setSource', source);
   }
 
+  setSearchAggregationId(uuid: string): void {
```
oshi97, line 51: should we just call it `setAutocompleteSessionId`? or did Rose have other thoughts

src/answers-headless.ts
```diff
@@ -48,6 +48,10 @@ export default class AnswersHeadless {
...
+    this.stateManager.dispatchEvent('query/setSearchAggregationId', uuid);
```
oshi97, line 52: if somebody wanted to clear the id would they just set it to blank string?

src/answers-headless.ts
yen-tt: nope, it just thought it would be a bit clearer in what autocompleteSessionId is doing. But I can just change it, not a problem

src/answers-headless.ts
yen-tt: yeah, do you think undefined/null is better?

oshi97: eh think it's fine!

PR #59, opened 2021-12-06
Add documentation for models

nmanu1:
Add TSDocs for models. Remove `sortBys` from `FiltersState` since it is also already present in `VerticalSearchState`. Add an ESLint plugin for TSDocs to give warnings for TSDoc formatting.

J=SLAP-1737
TEST=auto

Check that jest tests passed

### DIFF
--- a/src/models/slices/directanswer.ts
+++ b/src/models/slices/directanswer.ts
@@ -1,7 +1,7 @@
 import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
 
 /**
- * The state for a direct answer.
+ * Maintains the direct answer associated with the latest search.
  */
 export interface DirectAnswerState {
   /**
--- a/src/models/slices/directanswer.ts
+++ b/src/models/slices/directanswer.ts
@@ -1,5 +1,11 @@
 import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
 
+/**
+ * The state for a direct answer.
+ */
 export interface DirectAnswerState {
+  /**
+   * The data for the direct answer.
+   */
   result?: FeaturedSnippetDirectAnswer | FieldValueDirectAnswer;
 }
--- a/src/models/slices/directanswer.ts
+++ b/src/models/slices/directanswer.ts
@@ -5,8 +5,9 @@ import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answe
  */
 export interface DirectAnswerState {
   /**
-   * The data for the direct answer. The type of the data is based on where the
-   * answer was found by the Answers API.
+   * The data for the direct answer. The type of the data is determined by the
+   * Answers API based on whether the answer was found within a document or was a
+   * field value in the knowledge graph.
    */
   result?: FeaturedSnippetDirectAnswer | FieldValueDirectAnswer;
 }
--- a/src/models/slices/directanswer.ts
+++ b/src/models/slices/directanswer.ts
@@ -5,7 +5,8 @@ import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answe
  */
 export interface DirectAnswerState {
   /**
-   * The data for the direct answer.
+   * The data for the direct answer. The type of the data is based on where the
+   * answer was found by the Answers API.
    */
   result?: FeaturedSnippetDirectAnswer | FieldValueDirectAnswer;
 }
--- a/src/models/slices/filters.ts
+++ b/src/models/slices/filters.ts
@@ -1,8 +1,16 @@
 import { DisplayableFacet, SortBy } from '@yext/answers-core';
 import { SelectableFilter } from '../utils/selectablefilter';
 
+/**
+ * The state for filters.
+ */
 export interface FiltersState {
+  /**
+   * The array of static filters to apply to a search, with data on whether each filter is selected.
+   */
   static?: SelectableFilter[];
+  /**
+   * The array of facet filters to apply to a search, with data on how they should be displayed.
+   */
   facets?: DisplayableFacet[];
-  sortBys?: SortBy[]
 }
--- a/src/models/slices/filters.ts
+++ b/src/models/slices/filters.ts
@@ -1,16 +1,18 @@
-import { DisplayableFacet, SortBy } from '@yext/answers-core';
+import { DisplayableFacet } from '@yext/answers-core';
 import { SelectableFilter } from '../utils/selectablefilter';
 
 /**
- * Maintains the filters applied to the latest search.
+ * Maintains the current state of facets and filters in the application.
  */
 export interface FiltersState {
   /**
-   * The array of static filters, with data on whether each filter is selected.
+   * The collection of possible static filters that can be applied to the
+   * search results.
    */
   static?: SelectableFilter[];
   /**
-   * The array of facet filters, with data on how they should be displayed.
+   * The dynamic collection of facets that can be applied to filter the search
+   * results.
    */
   facets?: DisplayableFacet[];
 }
--- a/src/models/slices/filters.ts
+++ b/src/models/slices/filters.ts
@@ -7,12 +7,12 @@ import { SelectableFilter } from '../utils/selectablefilter';
 export interface FiltersState {
   /**
    * The collection of possible static filters that can be applied to the
-   * search results and whether they are currently selected.
+   * search results and whether each of them is currently selected.
    */
   static?: SelectableFilter[];
   /**
    * The dynamic collection of facets that can be applied to filter the search
-   * results and whether they are currently selected.
+   * results and whether each of them is currently selected.
    */
   facets?: DisplayableFacet[];
 }
--- a/src/models/slices/filters.ts
+++ b/src/models/slices/filters.ts
@@ -7,12 +7,12 @@ import { SelectableFilter } from '../utils/selectablefilter';
 export interface FiltersState {
   /**
    * The collection of possible static filters that can be applied to the
-   * search results.
+   * search results and whether they are currently selected.
    */
   static?: SelectableFilter[];
   /**
    * The dynamic collection of facets that can be applied to filter the search
-   * results.
+   * results and whether they are currently selected.
    */
   facets?: DisplayableFacet[];
 }
--- a/src/models/slices/filters.ts
+++ b/src/models/slices/filters.ts
@@ -2,15 +2,15 @@ import { DisplayableFacet, SortBy } from '@yext/answers-core';
 import { SelectableFilter } from '../utils/selectablefilter';
 
 /**
- * The state for filters.
+ * Maintains the filters applied to the latest search.
  */
 export interface FiltersState {
   /**
-   * The array of static filters to apply to a search, with data on whether each filter is selected.
+   * The array of static filters, with data on whether each filter is selected.
    */
   static?: SelectableFilter[];
   /**
-   * The array of facet filters to apply to a search, with data on how they should be displayed.
+   * The array of facet filters, with data on how they should be displayed.
    */
   facets?: DisplayableFacet[];
 }
--- a/src/models/slices/location.ts
+++ b/src/models/slices/location.ts
@@ -6,7 +6,8 @@ import { LatLong, LocationBias } from '@yext/answers-core';
  */
 export interface LocationState {
   /**
-   * The geographical location bias used in the search, from the Answers API.
+   * The geographical location bias used in the search, returned from the
+   * Answers API.
    */
   locationBias?: LocationBias,
   /**
--- a/src/models/slices/location.ts
+++ b/src/models/slices/location.ts
@@ -1,7 +1,8 @@
 import { LatLong, LocationBias } from '@yext/answers-core';
 
 /**
- * Maintains the user's location, if given, or the inferred location, that is used to bias search results.
+ * Maintains the user's location, if given, or the inferred location, that is
+ * used to bias search results.
  */
 export interface LocationState {
   /**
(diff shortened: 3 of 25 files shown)


src/models/slices/directanswer.ts
```diff
@@ -1,5 +1,11 @@
 import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
 
+/**
```
cea2aj, line 3: This often difficult to do, but we strive to write documentation that provides more information than the name itself. For example, `DirectAnswerState` could be documented as: 'Maintains the direct answer associated with the latest search'

src/models/slices/location.ts
```diff
@@ -1,6 +1,15 @@
 import { LatLong, LocationBias } from '@yext/answers-core';
 
+/**
+ * The state for the user's location.
+ */
 export interface LocationState {
+  /**
+   * The geographical location bias used in a search.
```
cea2aj, line 8: Can we make it clear that the `locationBias` is provided by the backend, and that `userLocation` affects search requests

nmanu1: Updated!

src/models/slices/directanswer.ts
nmanu1: Updated this one and some others to be more informative

src/models/state-manager.ts
```diff
@@ -2,8 +2,25 @@ import { Unsubscribe } from '@reduxjs/toolkit';
 import { State } from './state';
 import StateListener from './state-listener';
 
+/**
+ * Manages the information contained in the state for an AnswersHeadless instance.
```
tmeyer2115, line 6: I think somewhere in the docs for this class we should indicate that the `state` is immutable and can only be changed by dispatched events.

src/slices/filters.ts
```diff
@@ -27,9 +27,6 @@ const reducers = {
   setFacets: (state: FiltersState, action: PayloadAction<DisplayableFacet[]>) => {
     state.facets = action.payload;
   },
-  setSortBys: (state: FiltersState, action: PayloadAction<SortBy[]>) => {
```
tmeyer2115, line 30: I think we should keep sorts in here and remove them from the `vertical` state. The `filters` slice of state was meant to represent facets, filters, and sorts. @cea2aj or @oshi97, do you know why sorts appear in the `vertical` state chunk too?

src/models/slices/meta.ts
```diff
@@ -1,7 +1,20 @@
 import { Context } from '@yext/answers-core';
 
+/**
+ * Maintains the metadata for the latest search.
```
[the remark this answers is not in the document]
nmanu1: Oh ok. So, something like "Maintains the metadata for network requests"?

src/models/slices/directanswer.ts
```diff
@@ -1,5 +1,11 @@
...
+ * Maintains the direct answer associated with the latest search.
+ */
 export interface DirectAnswerState {
+  /**
+   * The data for the direct answer.
```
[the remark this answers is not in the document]
nmanu1: Something like "If found within a document, the result will be a `FeaturedSnippetDirectAnswer`. If it came from a field in the knowledge graph, the result will be a `FieldValueDirectAnswer`"?

src/models/slices/directanswer.ts
```diff
@@ -1,5 +1,11 @@
 import { FeaturedSnippetDirectAnswer, FieldValueDirectAnswer } from '@yext/answers-core';
 
+/**
+ * Maintains the direct answer associated with the latest search.
```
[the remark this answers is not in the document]
cea2aj: Yes, we should use them because that is our convention in Core. If I recall, they are inferred, but API extractor gives a warning if you don't include them

cea2aj:
Expanding on what Tom said, we follow the [TSDoc standard](https://tsdoc.org/) for our Typescript documentation. The biggest impact this has is marking public interfaces with the `@public` tag at the interface level. Additional tags which we commonly use include `@remarks`, `@example` and `@inheritDoc`. We also add a period to the end of each sentence because it looks better in the generated docs. I recommend looking through our Answers Core documentation to get a sense of how we adhere to the TSDoc standard

src/models/slices/directanswer.ts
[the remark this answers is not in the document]
nmanu1: I thought this might be good to do as part of the API extractor item if we needed to work out what should be `@public` and not. Unless, everything that is currently being exported should be `@public`, then I can just add that now as I add the docs.

src/slices/filters.ts
cea2aj: I'm not aware of any reason for them to be in both slices, and I agree that it should only be in one of them. I spoke with Nidhi about it, and she said that having the sorts in the filters slice was potentially confusing because it almost looks like the filters themselves are being sorted. We were thinking of keeping them in the `vertical` state to avoid that confusion

src/models/slices/directanswer.ts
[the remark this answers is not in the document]
cea2aj: That makes sense. There will almost certainly need to be some changes once we add API extractor, and it will tell us exactly what those changes need to be

src/models/state-manager.ts
nmanu1: I added this in `@remarks` for the interface

src/models/state-listener.ts
```diff
@@ -1,6 +1,19 @@
 import { State } from './state';
 
+/**
+ * Represents a listener for a value in the state.
+ */
 export default interface StateListener<T> {
+  /**
+   * Accesses a value in the state.
```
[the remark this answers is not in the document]
nmanu1: Updated here and for the interface description

src/models/slices/vertical.ts
```diff
@@ -8,26 +8,84 @@ import {
...
+     * {@inheritDoc AllResultsForVertical}
+     */
     allResultsForVertical: AllResultsForVertical,
+    /**
+     * The results from alternative verticals.
+     */
     alternativeVerticals: VerticalResults[],
   }
+  /**
+   * The number of results to skip.
```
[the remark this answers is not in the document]
nmanu1: Added more detail here

src/models/slices/sessiontracking.ts
```diff
@@ -1,4 +1,13 @@
+/**
+ * Maintains session information and whether it should be tracked.
```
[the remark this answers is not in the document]
nmanu1: Updated!

src/models/slices/filters.ts
```diff
@@ -1,8 +1,16 @@
 import { DisplayableFacet, SortBy } from '@yext/answers-core';
 import { SelectableFilter } from '../utils/selectablefilter';
 
+/**
+ * Maintains the filters applied to the latest search.
+ */
 export interface FiltersState {
+  /**
+   * The array of static filters, with data on whether each filter is selected.
```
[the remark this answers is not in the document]
nmanu1: Yes, that makes sense. Made some changes to static filters and facets and a few other places that were only giving redundant information.

src/models/slices/location.ts
```diff
@@ -1,6 +1,15 @@
 import { LatLong, LocationBias } from '@yext/answers-core';
 
+/**
+ * Maintains the user's location, if given, or the inferred location, that is used to bias search results.
+ */
 export interface LocationState {
+  /**
+   * The geographical location bias used in the search, from the Answers API.
```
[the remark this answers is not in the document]
nmanu1: Yes, it is. Updated

src/models/slices/directanswer.ts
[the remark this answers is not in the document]
cea2aj: Maybe it could indicate that the type depends on the search configuration and what the Answers algorithms determines?

src/models/slices/filters.ts
```diff
@@ -1,8 +1,18 @@
-import { DisplayableFacet, SortBy } from '@yext/answers-core';
+import { DisplayableFacet } from '@yext/answers-core';
 import { SelectableFilter } from '../utils/selectablefilter';
 
+/**
+ * Maintains the current state of facets and filters in the application.
+ */
 export interface FiltersState {
+  /**
+   * The collection of possible static filters that can be applied to the
```
cea2aj, line 9: Can we also indicate that it maintains which filters are currently applied?

nmanu1: Sure! This should be added for both static filters and facets, right?

src/models/slices/query.ts
```diff
@@ -1,10 +1,32 @@
 import { QuerySource, QueryTrigger, SearchIntent } from '@yext/answers-core';
 
+/**
+ * Maintains the latest query and its associated data.
+ */
 export interface QueryState {
+  /**
+   * The latest user query input.
```
cea2aj, line 8: What do you think about wording this as "The input used in the next search"

nmanu1: I think it would be good to still specify that it's user input. Maybe "The user input for the next query to search"?

cea2aj: Maybe "The user input used for the next search query"?

src/models/slices/filters.ts
cea2aj: Yep

src/models/slices/query.ts
nmanu1: Yes, that works!

src/models/slices/directanswer.ts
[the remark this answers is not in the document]
nmanu1: Like this? "The type of the data is based on where the answer was found by the Answers API."

src/models/slices/filters.ts
nmanu1: Updated

src/models/slices/query.ts
```diff
@@ -1,10 +1,32 @@
...
+  /**
+   * The user input used for the next search query.
+   */
   input?: string,
+  /**
+   * The ID of the query from the latest search.
+   */
   queryId?: string,
+  /**
+   * How the query was triggered (besides user input).
```
oshi97, line 16: would we want to use `@inheritDoc` for any of these? Not sure if it works, but maybe it could simplify some things?

tests/utils/setTimeout.ts
```diff
@@ -2,7 +2,7 @@ import timers from 'timers';
 
 /**
  * Promise version of setTimeout. timers/promises was only added in node 16
- * @param ms time in ms to wait
+ * @param ms - time in ms to wait
```
oshi97, line 5: we could try using the tsdoc eslint plugin? I think there's some tsdoc-param-tag-missing-hyphen rule?
you can see it if you remove a hyphen from the example at https://microsoft.github.io/tsdoc/

nmanu1: sure, I can add that

src/models/slices/query.ts
nmanu1: yes, I've been using `@inheritDoc` in some places. I didn't use it for `queryTrigger` because the current description felt a little more direct/simple to me. but, if we want the extra detail, I can use `{@inheritDoc QueryTrigger}` instead

src/models/slices/meta.ts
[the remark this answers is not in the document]
nmanu1: Updated

tests/utils/setTimeout.ts
nmanu1: added the eslint rule so that it now gives warnings for tsdoc formatting

src/models/slices/directanswer.ts
```diff
@@ -1,5 +1,12 @@
...
+ */
 export interface DirectAnswerState {
+  /**
+   * The data for the direct answer. The type of the data is based on where the
+   * answer was found by the Answers API.
```
tmeyer2115, line 9: I think we should provide a bit more detail on the possible sources of data here.

src/models/slices/filters.ts
```diff
@@ -1,8 +1,18 @@
...
+   * search results and whether they are currently selected.
```
tmeyer2115, line 10: Minor nit: Here and below, the phrase "whether they are currently selected" is a bit misleading. It makes it sound as if the whole array is selected or unselected. It may be more accurate to say "and which are currently selected".

src/models/slices/meta.ts
```diff
@@ -1,7 +1,20 @@
 import { Context } from '@yext/answers-core';
 
+/**
+ * Maintains the metadata for network requests.
+ */
 export interface MetaState {
+  /**
+   * A JSON object used for triggering and passing data to query rules.
```
tmeyer2115, line 8: It might be helpful here to link to the official docs around `Context`. I think there is a page in the API documentation somewhere that explains what it is and the limitations. Rose could point you in the right direction.

src/models/state-listener.ts
```diff
@@ -1,18 +1,19 @@
 import { State } from './state';
 
 /**
- * Represents a listener for a value in the state.
+ * Represents a listener for a value of type T in the state.
```
tmeyer2115, line 4: Minor nit: This phrasing could be misinterpreted as "listens for all values of type T in the state"

src/models/slices/directanswer.ts
nmanu1: Something like this? "The type of the data is determined by the Answers API based on whether the answer was found within a document or was a field value in the knowledge graph."

src/models/slices/meta.ts
nmanu1: Added!

src/models/slices/filters.ts
tmeyer2115: Sure that sounds good

src/models/slices/directanswer.ts
tmeyer2115: Sounds good!

nmanu1: Updated!

src/models/slices/filters.ts
nmanu1: Updated

tmeyer2115 (approved):

PR #60, opened 2021-12-07
Add documentation for slices and utils

nmanu1:
Update documentation for `slices` and `utils` in the `src` folder.

J=SLAP-1737
TEST=none

### DIFF
--- a/src/slices/directanswer.ts
+++ b/src/slices/directanswer.ts
@@ -15,10 +15,6 @@ const reducers = {
 
 /**
  * Registers with Redux the slice of {@link State} pertaining to direct answers.
- *
- * @param prefix - The prefix to access the part of the Redux store specific to the
- *                 AnswersHeadless instance
- * @returns The {@link Slice} for direct answers
  */
 export default function createDirectAnswerSlice(prefix: string): Slice<DirectAnswerState, typeof reducers> {
   return createSlice({
--- a/src/slices/directanswer.ts
+++ b/src/slices/directanswer.ts
@@ -16,7 +16,8 @@ const reducers = {
 /**
  * Registers with Redux the slice of {@link State} pertaining to direct answers.
  *
- * @param prefix - The prefix for the AnswersHeadless instance
+ * @param prefix - The prefix to access the part of the Redux store specific to the
+ *                 AnswersHeadless instance
  * @returns The {@link Slice} for direct answers
  */
 export default function createDirectAnswerSlice(prefix: string): Slice<DirectAnswerState, typeof reducers> {
--- a/src/slices/directanswer.ts
+++ b/src/slices/directanswer.ts
@@ -15,6 +15,9 @@ const reducers = {
 
 /**
  * Registers with Redux the slice of {@link State} pertaining to direct answers.
+ *
+ * @param prefix - The prefix for the AnswersHeadless instance
+ * @returns The {@link Slice} for direct answers
  */
 export default function createDirectAnswerSlice(prefix: string): Slice<DirectAnswerState, typeof reducers> {
   return createSlice({
--- a/src/slices/filters.ts
+++ b/src/slices/filters.ts
@@ -1,5 +1,5 @@
 import { createSlice, PayloadAction, Slice } from '@reduxjs/toolkit';
-import { Filter, FacetOption, DisplayableFacet, SortBy } from '@yext/answers-core';
+import { Filter, FacetOption, DisplayableFacet } from '@yext/answers-core';
 import { FiltersState } from '../models/slices/filters';
 import { SelectableFilter } from '../models/utils/selectablefilter';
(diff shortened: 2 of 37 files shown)


src/utils/transform-filters.ts
```diff
@@ -15,8 +20,11 @@ function combineFiltersWithOR(filters: Filter[]): Filter | CombinedFilter {
 }
 
 /**
- * Convert a list of SelectableFilters use in Answers Headless
- * to a single nested filter stucture use in Answers Core
+ * Converts a list of {@link SelectableFilter}s used in Answers Headless to a
+ * single nested filter stucture used in Answers Core.
+ *
+ * @param selectableFilters - The filters to be transformed
+ * @returns The filters in a singly-nested structure
```
yen-tt, line 27: can we link Filter and CombinedFilter here for the return structure?

src/slices/directanswer.ts
```diff
@@ -15,6 +15,9 @@ const reducers = {
 
 /**
  * Registers with Redux the slice of {@link State} pertaining to direct answers.
+ *
+ * @param prefix - The prefix for the AnswersHeadless instance
```
yen-tt, line 19: maybe we can be more descriptive for `prefix` to inform that it's an answers headless id to access specific chunk of data in redux

nmanu1: something like this? "The prefix to access the part of the Redux store specific to the AnswersHeadless instance"

src/utils/filter-creators.ts
```diff
@@ -1,14 +1,18 @@
 import { CombinedFilter, Filter, FilterCombinator, Matcher, NearFilterValue } from '@yext/answers-core';
 import { Range } from '../models/utils/range';
 
+/**
+ * A union type for the different kinds of filter.
+ */
 type filterTypes = Filter | CombinedFilter;
```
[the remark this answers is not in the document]
nmanu1: sure!

src/utils/transform-filters.ts
nmanu1: updated!

src/slices/directanswer.ts
yen-tt: sounds good

src/slices/directanswer.ts
```diff
@@ -15,6 +15,10 @@ const reducers = {
 
 /**
  * Registers with Redux the slice of {@link State} pertaining to direct answers.
+ *
+ * @param prefix - The prefix to access the part of the Redux store specific to the
+ *                 AnswersHeadless instance
```
cea2aj, line 20: @oshi97 Can you confirm that this is accurate?

src/slices/filters.ts
```diff
@@ -6,14 +6,35 @@ import { areFiltersEqual } from '../utils/filter-utils';
 
 const initialState: FiltersState = {};
 
+/**
+ * The payload for updating a facet option.
```
cea2aj, line 10: @tmeyer2115 Do we want full internal documentation as well? I was under the impression the documentation for this item was focused on the public part of the API. Many of our internal models are well explained by their names, and I am dubious of the benefits of maintaining full internal documentation. Clean Code recommends avoiding mandated internal javadocs

tmeyer2115: I'm in favor of some internal docs. If a method or model is adequately explained by its name, I'd say we don't need JSDocs. But, if we feel some additional clarification is needed, especially for internal classes or methods, I think we should add internal docs.

tmeyer2115: Obviously, the ideal is what Fancher would call "self-documenting code", but that's not always possible for a method or class of suitable complexity.

tmeyer2115: @cea2aj and @nmanu1, given the time constraints we're up against, I'm fine if we hold off on any internal docs. Getting API-Extractor set up and all the public facing things documented is more important.

nmanu1: Sounds good. I've removed the redundant internal docs and will hold off on adding any more

src/slices/directanswer.ts
oshi97: Yup!

src/slices/directanswer.ts
```diff
@@ -15,6 +15,10 @@ const reducers = {
...
+ * @returns The {@link Slice} for direct answers
```
oshi97, line 21: is it possible to have these slice creators have some kind of shared type? Then maybe we can consolidate this documentation instead of having it copied for every slice creator we have

oshi97 (commented):
overall looks really good!

src/slices/directanswer.ts
nmanu1: ended up removing these because the slice creators aren't exported to API extractor

oshi97 (approved):

PR #61, opened 2021-12-09
Add remaining documentation

nmanu1:
Add TSDocs for the remaining files in the `src` folder that are exported.

J=SLAP-1737
TEST=none

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -158,8 +158,8 @@ export default class AnswersHeadless {
   }
 
   /**
-   * Sets {@link VerticalSearchState.alternativeVerticals} to the specified
-   * verticals.
+   * Sets the alternativeVerticals for {@link VerticalSearchState.noResults} to the
+   * specified verticals.
    *
    * @param alternativeVerticals - The alternative verticals to set
    */
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
 import { SelectableFilter } from './models/utils/selectablefilter';
 import { transformFiltersToCoreFormat } from './utils/transform-filters';
 
+/**
+ * Provides methods for updating the state and for sending requests to the Answers
+ * API through {@link AnswersCore}.
+ */
 export default class AnswersHeadless {
+  /**
+   * Common utility functions for manipulating Answers-related data.
+   */
   public readonly utilities = answersUtilities;
 
+  /**
+   * Sets up the AnswersHeadless instance with managers for its state and network
+   * requests and an instance of the {@link AnswersCore} library.
+   *
+   * @param core - The Answers Core instance used for sending network requests
+   * @param stateManager - The manager of the state and how it can be manipulated
+   * @param httpManager - The manager for network requests and reponses through core
+   */
   constructor(
     private core: AnswersCore,
     private stateManager: StateManager,
     private httpManager: HttpManager,
   ) {}
 
+  /**
+   * Sets {@link QueryState.input} to the specified input.
+   *
+   * @param input - The new query input to update to
+   */
   setQuery(input: string): void {
     this.stateManager.dispatchEvent('query/setInput', input);
   }
 
+  /**
+   * Sets {@link QueryState.queryTrigger} to the specified trigger.
+   *
+   * @param trigger - The new query trigger to update to
+   */
   setQueryTrigger(trigger: QueryTrigger): void {
     this.stateManager.dispatchEvent('query/setTrigger', trigger);
   }
 
+  /**
+   * Sets {@link QueryState.querySource} to the specified source.
+   *
+   * @param source - The new query source to update to
+   */
   setQuerySource(source: QuerySource): void {
     this.stateManager.dispatchEvent('query/setSource', source);
   }
 
+  /**
+   * Sets {@link VerticalSearchState.verticalKey} to the specified key.
+   *
+   * @param verticalKey - The new vertical key to update to
+   */
   setVerticalKey(verticalKey: string): void {
     this.stateManager.dispatchEvent('vertical/setVerticalKey', verticalKey);
   }
 
+  /**
+   * Sets {@link VerticalSearchState.limit} to the specified limit.
+   *
+   * @param limit - The new vertical limit to update to
+   */
   setVerticalLimit(limit: number): void {
     this.stateManager.dispatchEvent('vertical/setLimit', limit);
   }
 
+  /**
+   * Sets {@link UniversalSearchState.limit} to the specified limit.
+   *
+   * @param limit - The new universal limit to update to
+   */
   setUniversalLimit(limit: UniversalLimit): void {
     this.stateManager.dispatchEvent('universal/setLimit', limit);
   }
 
+  /**
+   * Sets {@link VerticalSearchState.offset} to the specified offset.
+   *
+   * @param offset - The new vertical offset to update to
+   */
   setOffset(offset: number): void {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
+  /**
+   * Sets {@link FiltersState."static"} to the specified filters.
+   *
+   * @param filters - The new static filters to update to
+   */
   setStaticFilters(filters: SelectableFilter[]): void {
     this.stateManager.dispatchEvent('filters/setStatic', filters);
   }
 
+  /**
+   * Sets {@link FiltersState.facets} to the specified facets.
+   *
+   * @param facets - The new facets to update to
+   */
   setFacets(facets: DisplayableFacet[]): void {
     this.stateManager.dispatchEvent('filters/setFacets', facets);
   }
 
+  /**
+   * Resets all {@link FiltersState.facets | facets} to be unselected.
+   */
   resetFacets(): void {
     this.stateManager.dispatchEvent('filters/resetFacets');
   }
 
+  /**
+   * Sets {@link SpellCheckState.enabled} to the specified boolean value.
+   *
+   * @param enabled - Whether spellcheck should be set to enabled or not
+   */
   setSpellCheckEnabled(enabled: boolean): void {
     this.stateManager.dispatchEvent('spellCheck/setEnabled', enabled);
   }
 
+  /**
+   * Sets {@link SessionTrackingState.enabled} to the specified boolean value.
+   *
+   * @param enabled - Whether session tracking should be set to enabled or not
+   */
   setSessionTrackingEnabled(enabled: boolean): void {
     this.stateManager.dispatchEvent('sessionTracking/setEnabled', enabled);
   }
 
+  /**
+   * Sets {@link SessionTrackingState.sessionId} to the specified ID.
+   *
+   * @param sessionId - The new session ID to update to
+   */
   setSessionId(sessionId: string): void {
     this.stateManager.dispatchEvent('sessionTracking/setSessionId', sessionId);
   }
 
+  /**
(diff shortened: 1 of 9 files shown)


src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
 import { SelectableFilter } from './models/utils/selectablefilter';
 import { transformFiltersToCoreFormat } from './utils/transform-filters';
 
+/**
+ * Provides methods for updating the state and for sending requests to the Answers
+ * API through {@link AnswersCore}.
```
cea2aj, line 32: I don' t think consumers of Headless need to know about Answers Core. What do you think about something like: "Provides the functionality for interacting with an Answers Search experience"

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
   public readonly utilities = answersUtilities;
 
+  /**
+   * Sets up the AnswersHeadless instance with managers for its state and network
+   * requests and an instance of the {@link AnswersCore} library.
+   *
+   * @param core - The Answers Core instance used for sending network requests
+   * @param stateManager - The manager of the state and how it can be manipulated
+   * @param httpManager - The manager for network requests and reponses through core
+   */
```
cea2aj, line 47: This constructor is actually internal since users of the library will be able to get a Headless instance through the `provideAnswersHeadless` function.  If you take a look at the `AnswersCore` class, you'll see we mark it as internal: https://github.com/yext/answers-core/blob/master/src/AnswersCore.ts

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
   constructor(
     private core: AnswersCore,
     private stateManager: StateManager,
     private httpManager: HttpManager,
   ) {}
 
+  /**
+   * Sets {@link QueryState.input} to the specified input.
+   *
+   * @param input - The new query input to update to
```
cea2aj, line 57: What do you think about the wording "The input which is set" instead?

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
+   * Sets {@link VerticalSearchState.offset} to the specified offset.
+   *
+   * @param offset - The new vertical offset to update to
+   */
   setOffset(offset: number): void {
     this.stateManager.dispatchEvent('vertical/setOffset', offset);
   }
 
+  /**
+   * Sets {@link FiltersState."static"} to the specified filters.
```
cea2aj, line 118: Are quotes needed around "static" here?

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
+   * Sets {@link FiltersState.facets} to the specified facets.
+   *
+   * @param facets - The new facets to update to
+   */
   setFacets(facets: DisplayableFacet[]): void {
     this.stateManager.dispatchEvent('filters/setFacets', facets);
   }
 
+  /**
+   * Resets all {@link FiltersState.facets | facets} to be unselected.
```
cea2aj, line 136: Thoughts on "Unselects all {@link FilterState.facets | facets}"?

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
+   */
   resetFacets(): void {
     this.stateManager.dispatchEvent('filters/resetFacets');
   }
 
+  /**
+   * Sets {@link SpellCheckState.enabled} to the specified boolean value.
+   *
+   * @param enabled - Whether spellcheck should be set to enabled or not
```
cea2aj, line 145: nit: I have a small preference for putting "Whether or not" at the front of the phrase, but up to you!

src/answers-headless.ts
```diff
@@ -27,108 +27,247 @@ import answersUtilities from './answers-utilities';
...
+   * @param request - The data for the network request
+   * @returns A Promise of a {@link QuestionSubmissionResponse} from the Answers API
+   */
   async submitQuestion(request: QuestionSubmissionRequest): Promise<QuestionSubmissionResponse> {
     return this.core.submitQuestion(request);
   }
 
-  async executeUniversalQuery(): Promise<UniversalSearchResponse | undefined> {
+  /**
+   * Uses {@link AnswersCore} to perform an Answers search across all verticals with
```
cea2aj, line 264: Again, I don't think we need to indicate that `AnswersCore` is being used here. How about something like: "Performs an answers search across all verticals with relevant parts of the state used as input to the search"?

src/answers-headless.ts
```diff
@@ -255,6 +415,17 @@ export default class AnswersHeadless {
     });
   }
 
+  /**
+   * Uses {@link AnswersCore} to perform a filtersearch request against specified
+   * fields within a single vertical with the specified parameters and the
```
cea2aj, line 420: I don't think we need the "with the specified parameters" part, however it might be good to mention that the verticalKey comes from the state.

src/answers-utilities.ts
```diff
@@ -2,6 +2,15 @@ import { DisplayableFacet } from '@yext/answers-core';
 import { isLevenshteinMatch } from './utils/searchable-facets';
 
 export default {
+  /**
+   * Searches through the specified facet and filters out the options that aren't a
+   * Levenshtein match for the given searchTerm.
```
cea2aj, line 7: I'm not sure that we need to include the matching algorithm (Levenshtein) is these docs

src/http-manager.ts
```diff
@@ -5,16 +5,25 @@ type ServiceIds = {
...
 /**
- * Assign numeric ids to every http requests and the corresponding responses
+ * Assigns numeric IDs to every http request and the corresponding response
  * through {@link AnswersCore}. This helps track the received order of requests
- * and responses. {@link AnswersHeadless} use it to ensure dispatch event for
- * state update is trigger from up-to-date responses (e.g. if the new received
- * response have higher id number from the recorded received response)
+ * and responses. {@link AnswersHeadless} uses it to ensure dispatch events for
+ * state updates are triggered with up-to-date responses (e.g. if the newly received
+ * response has a higher ID number than the recorded received response).
```
cea2aj, line 12: I believe this class is internal

src/answers-headless.ts
nmanu1: Ah ok, I'll remove these docs then

src/headless-reducer-manager.ts
```diff
@@ -17,12 +17,29 @@ import createSearchStatusSlice from './slices/searchstatus';
  * Manages the current map of headless IDs to Reducers.
  */
 export default class HeadlessReducerManager {
+  /**
```
cea2aj, line 20: I believe this class is internal as well

src/answers-headless.ts
nmanu1: Sure, I can change it to "The input to set" (and similar for the others)

src/answers-headless.ts
nmanu1: That works!

src/headless-reducer-manager.ts
```diff
@@ -17,12 +17,29 @@ import createSearchStatusSlice from './slices/searchstatus';
...
   addAnswersReducer(headlessId: string): void {
     this.headlessIdToReducer[headlessId] = createAnswersReducer(headlessId + '/');
   }
 
+  /**
+   * Gets the parent Reducer for the {@link ParentState} of the mapping of
+   * AnswersHeadless instances.
+   *
+   * @returns A parent Reducer function that is a combination of the Reducers in
+   *          {@link headlessIdToReducer}
```
cea2aj, line 41: I'm not sure that this link will work. I would expect that it would be {@link HeadlessReducerManager.headlessIdToReducer}. We could figure that out once you add API extractor, however since this class is internal, I don't think it will process it

src/http-manager.ts
nmanu1: Yes, I'll remove some docs I added for this class, `ReduxStateManager`, `HttpManager`, and others that are internal

src/headless-reducer-manager.ts
nmanu1: Removed this since it's internal

src/headless-reducer-manager.ts
nmanu1: Removed the new docs because it's internal

src/answers-headless.ts
nmanu1: That makes sense. I'll change it to that

src/answers-headless.ts
nmanu1: Updated this and the others similarly

src/index.ts
```diff
@@ -10,7 +10,13 @@ import { SessionTrackingState } from './models/slices/sessiontracking';
...
 
-type HeadlessConfig = AnswersConfig & {
+/**
+ * The configuration for an AnswersHeadless instance.
+ */
+export type HeadlessConfig = AnswersConfig & {
+  /**
+   * The ID of the AnswersHeadless instance.
+   */
   headlessId?: string
```
yen-tt, line 20: Maybe we should add a remark that this id cannot be the same as DEFAULT_HEADLESS_ID?

nmanu1: Added! I didn't reference DEFAULT_HEADLESS_ID in the remark directly because we aren't planning on exporting that constant.

yen-tt: hmm I feel like if we don't reference it, then it's a little vague and user wouldn't really know what can't be use. If we do want to add this remark, I think it would be better to export that variable and mark it with @/internal. @cea2aj what do you think?

cea2aj: I agree that it would be good to include the value of the variable which is reserved. We could either mention that the reserved ID is 'main'. There is a risk that we update the ID in the constants file and forget to update it here though. For that reason, I'm in favor of exporting that constant and linking to it. We can't mark it as internal or else API extractors wouldn't include a doc page for it. I think it's reasonable to consider it a public part of the interface since changes to the value affects which headlessIds can be used by consumers of the library

yen-tt: ah i didn't realize internal tag prevent api extractor from creating a doc page. I agree, yea let's export the constant and link it in the headless config

tmeyer2115: I'll preface this by saying I don't feel too strongly either way. In my opinion, we don't need to export the `DEFAULT_HEADLESS_ID`. Linking to it feels like overkill a bit. A reader of the documentation needs to know what the reserved value is. By linking to the variable itself, wouldn't we be making them do an extra hop to get to that info? They'd need to click into the link.

Internally, we know that we can't change this identifier already. I'm not worried about someone attempting that.

oshi97: can we change this from a string type to maybe some typescript type that is "any string but our special string"

nmanu1: I think it would be good to either make the constant public and link to it in the docs, or explicitly say in the docs what the value of the reserved ID is (although this may be a problem if we update the constant and forget to update the docs as Connor mentioned). Even if we made a type to check that, the user would still need to know which value is reserved by exporting the constant, right?

oshi97: yeah I'm cool with exporting it! I think the excluded string type is actually a bit harder than I thought so we can skip for now
https://stackoverflow.com/questions/51442157/type-for-every-possible-string-value-except

nmanu1: Updated to link directly to DEFAULT_HEADLESS_ID, which I will export in the API extractor PR

cea2aj (approved):

PR #63, opened 2021-12-10
Add API Extractor

nmanu1:
Add API extractor to get an API Report, a d.ts bundle, and markdown docs, as in Answers Core and Jambo. Also, add more exports, release tags, and link to markdown docs in README. Rename `Range` to `BoundedRange` to avoid conflict with the Typescript `Range` interface.

Notes:
- When reviewing, you can ignore the docs files
- As in Answers Core, there are separate `build-ci` and `api-extractor-ci` scripts, which exit with non-zero exit codes for warning and outdated api-reports
- As in Answers Core, the automatic tests will error out if the docs are out of date
- @yext/answers-core is added as a bundled package in the config for API extractor to correctly resolve links. This adds types/interfaces from answers-core in the d.ts bundle so that the referenced docs are not treated as coming from an external package

J=SLAP-1738
TEST=manual

Ran the api extractor and inspected the api report, type bundle, and markdown docs. Made changes to the docs and checked that the docs validation script worked as expected.


api-extractor.json
```diff
@@ -0,0 +1,364 @@
+/**
```
oshi97, line 1: just out of curiosity are there differences between this file and the api-extractor.json in answers-core? I remember you or Connor mentioning something about a config option "just working"

src/answers-core-re-exports.ts
```diff
@@ -0,0 +1,49 @@
+export {
```
oshi97, line 1: I guess api extractor doesn't work with `export *`?

api-extractor.json
nmanu1: yes, the difference is primarily the `bundledPackages`. here, we are adding answers-core as a bundled package, which adds it directly to the generated bundle for headless. this solves the problem of links not resolving because they are for types/interfaces in an external package

src/answers-core-re-exports.ts
nmanu1: API extractor does work with `export *`, but that would add everything exported from answers-core into the bundle for Headless. It seemed a little odd to have things like `provideCore()` exported when it isn't used with Headless, so this file exports only the Core types/interfaces that are actually used in Headless

README.md
```diff
@@ -1,5 +1,7 @@
 # answers-headless
-A library for powering UI components for Yext Answers integrations
+A library for powering UI components for Yext Answers integrations.
+
+[Documentation for Exported Types](./docs/answers-headless.md)
```
cea2aj, line 4: What do you think of renaming this to "Full Documentation"? Not everything that we export is a Type

src/answers-core-re-exports.ts
```diff
@@ -0,0 +1,49 @@
...
+  AnswersConfig,
+  AnswersCore,
```
cea2aj, line 3: Is there a reason we are exporting AnswersCore? I believe it's just an internal dependency of Answers Headless which doesn't need to be available publicly

etc/answers-headless.api.md
```diff
@@ -0,0 +1,670 @@
...
+import { Unsubscribe } from '@reduxjs/toolkit';
+
+// @public
+export interface AllResultsForVertical {
+    facets: DisplayableFacet[];
+    results: Result[];
+    resultsCount: number;
+}
+
+// Warning: (ae-forgotten-export) The symbol "AnswersConfigWithApiKey" needs to be exported by the entry point index.d.ts
```
cea2aj, line 16: Nidhi, can you take a look through this API report and add the exports that API extractors says is missing?  The one exception is the AnswersCore exports because I don't think that needs to be part of the public interface

src/answers-core-re-exports.ts
oshi97: sgtm!

scripts/validate-docs.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This script verifies that the docs are up to date with the source code
+# The script should be ran after new docs are built
+# If there are any git diffs after new docs are generated, the docs are out of date
+
+git diff --exit-code > /dev/null # send stdout to /dev/null to reduce clutter in the CI output
```
cea2aj, line 7: Can we update this to just check the diff on the docs folder?

README.md
nmanu1: Would "Full Documentation" imply that we're exporting and documenting everything in the repo?

src/answers-core-re-exports.ts
nmanu1: AnswersCore and BaseAnswersConfig are being exported because AnswersConfig links to them in its documentation. If they aren't exported, API extractor gives "unresolved link" warnings

README.md
cea2aj: I would expect a "Full Documentation" link of repo to provide documentation of the public interface. "Full Documentation" is also the wording we use in Answers Core and it's good to be consistent

etc/answers-headless.api.md
nmanu1: I looked through it, and if those are added, the only things we wouldn't be exporting from Core would be `AnswersCore` and `provideCore()`. (Also, see previous comment on why `AnswersCore` is exported). If that's what we want, to have full documentation for anything we export from Core, maybe there's an easier way to export everything except for one or two things?

README.md
nmanu1: Ok! I'll change it to that

src/answers-core-re-exports.ts
cea2aj: If you don't export AnswersCore, will the docs generate with just the "AnswersCore" text, or will it include a broken link?

src/models/index.ts
```diff
@@ -0,0 +1,23 @@
+import StateListener from './state-listener';
```
cea2aj, line 1: Rather than having a separate import and export, can these just be export statements like the other lines?

src/answers-core-re-exports.ts
nmanu1: It removes the text entirely, so the docs say something like "The main configuration options for .", where it should say "AnswersCore" before the period with a link

cea2aj: Ideally we wouldn't export `AnswersCore`, but since it breaks the docs, I think we should leave it in. I suppose we could remove the link from Answers Core so that we no longer need to export it, but that's out of scope

etc/answers-headless.api.md
cea2aj: For now I'd say it's okay to keep the list since we don't want `provideCore` exported. I experimented with `import { * as AnswersCoreExports } from '@yext/answers-core'` and attempting to delete single exports but I couldn't get it working. I don't think we should spend too much time on it now, and we can make a follow up item if needed

nmanu1: Updated to include all of the "forgotten exports" from Core

cea2aj (approved):

oshi97 (approved):

PR #68, opened 2021-12-14
v0.9 package update

yen-tt:
update version to 0.9 for EA release

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -1,7 +1,7 @@
 The following NPM packages may be included in this product:
 
- - @babel/runtime-corejs3@7.15.4
- - @babel/runtime@7.14.6
+ - @babel/runtime-corejs3@7.16.5
+ - @babel/runtime@7.16.5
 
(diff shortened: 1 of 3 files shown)


THIRD-PARTY-NOTICES
```diff
@@ -32,7 +32,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The following NPM package may be included in this product:
 
- - @reduxjs/toolkit@1.6.0
+ - @reduxjs/toolkit@1.7.0
```
oshi97, line 35: Should we update the version in the package.json to this?
Seeing this made me think we may as well just update all our deps? only redux-thunk and redux-toolkit are non-dev dependencies
image

oshi97: this is the package I used https://www.npmjs.com/package/npm-check-updates

yen-tt: updated!

nmanu1 (approved):

PR #69, opened 2021-12-16
Manage current vertical or universal search in state

cea2aj:
Enforce management of either vertical or universal search by headless.

This allows us to store whether or not headless is managing universal or vertical search, and our components can use that rather than the `isVertical` props. In the future we can update these methods to clear the state as necessary so that results or filters aren't left over when switching between vertical and universal.

The interface change:
- Add meta.searchType: SearchTypeEnum.Universal | SearchTypeEnum.Vertical
- Create a setUniversal() which sets the verticalKey to 'undefined' and sets the meta.searchType to 'universal'
- Rename setVerticalKey() to setVertical() and update it to set the meta.searchType to 'vertical'
- Allow an apiKey to be passed into the headless config which allows the searchType to be set when the headless instance is created

I copied the `EnumOrLiteral` type from the answers-analytics library which allows someone to supply either the enum value or the associated string value for the search type.

J=SLAP-1757
TEST=auto

Add relevant unit test and update other tests


src/answers-headless.ts
```diff
@@ -261,7 +271,12 @@ export default class AnswersHeadless {
    *
    * @returns A Promise of a {@link UniversalSearchResponse} from the Answers API
    */
-  async executeUniversalQuery(): Promise<UniversalSearchResponse> {
+  async executeUniversalQuery(): Promise<UniversalSearchResponse | undefined> {
```
yen-tt, line 274: I know that AlternativeVertical component in vertical page uses executeUniversalQuery to get results from all verticals, so it would have to switch between searchTypes from Vertical => Universal (execute search) => Vertical right? This would require us to pass a field, verticalKey, around for any component that requires this crossover logic. But, I think this is the only spot that cross between universal and vertical searches so this should be ok? just double checking

src/answers-headless.ts
```diff
@@ -75,12 +76,21 @@ export default class AnswersHeadless {
...
   /**
-   * Sets {@link VerticalSearchState.verticalKey} to the specified key.
+   * Sets up Headless to manage the vertical indicated by the verticalKey.
    *
    * @param verticalKey - The vertical key to set
    */
-  setVerticalKey(verticalKey: string): void {
+  setVertical(verticalKey: string): void {
     this.stateManager.dispatchEvent('vertical/setVerticalKey', verticalKey);
+    this.stateManager.dispatchEvent('meta/setSearchType', SearchTypeEnum.Vertical);
```
yen-tt, line 85: You mentioned before that we can derive whether or not it's universal/vertical search based on verticalKey. so if setVertical and setUniversal properly update verticalKey, do we still need SearchTypeEnum?

src/answers-headless.ts
cea2aj: AlternativeVerticals uses the `alternativeVerticals` and the `allResultsForVertical` data which are available from a vertical query. I don't believe it uses `executeUniversalQuery`

src/answers-headless.ts
cea2aj: Yes, we could just use the verticalKey for this, however this function allows us to add additional functionality such as clearing the state when `setVertical` or `setUniversal` is called. Currently we handle that in the sample app, however product has mentioned it might make sense in Headless as well. That functionality isn't part of this PR, but we may add it in the future. Clearing parts of the state is outside the scope of `setVerticalKey`, but it would make sense in `setVertical`. I also think that checking the searchType is a little clearer than checking whether or not the verticalKey exists

src/answers-headless.ts
yen-tt: oh I saw in the code where it calls `renderUniversalDetails` which calls `actions.executeUniversalQuery`. Is that something else? Or maybe it's not updated?

cea2aj: Oh, you are right. It does fire off a universal search, how that was because we didn't have a universal results component or the page linking figured out at that time. I think we should update that component so that clicking on the links works by going through the react router, and the page navigated to should fire off the search. So universal search inside Alternative verticals shouldn't be an issue after those updates

yen-tt (approved):

src/answers-headless.ts
```diff
@@ -261,7 +271,12 @@ export default class AnswersHeadless {
    *
    * @returns A Promise of a {@link UniversalSearchResponse} from the Answers API
    */
-  async executeUniversalQuery(): Promise<UniversalSearchResponse> {
+  async executeUniversalQuery(): Promise<UniversalSearchResponse | undefined> {
+    if(this.state.meta.searchType !== SearchTypeEnum.Universal) {
```
nmanu1, line 275: Does `executeUniversalAutocomplete` also need to have this console error added?

src/index.ts
```diff
@@ -26,7 +26,12 @@ export type HeadlessConfig = AnswersConfig & {
    * @remarks
    * Must be different from {@link DEFAULT_HEADLESS_ID}.
    */
-  headlessId?: string
+  headlessId?: string,
+  /**
+   * The verticalKey associated with the vertical to manage. If none is provided,
+   * Answers Headless will manage universal search
```
nmanu1, line 32: missing period at the end

src/models/utils/searchType.ts
```diff
@@ -0,0 +1,24 @@
+import { EnumOrLiteral } from '../../utils/types';
+
+/**
+ * An enum which indicates the type of search that headless is managing.
```
nmanu1, line 4: capitalize "headless" here and on lines 10 and 14

src/utils/types.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Produces a union type from the enum passed as a generic which consists of the enum values
+ * and the string literals of the enum
```
nmanu1, line 3: missing period at the end

etc/answers-headless.api.md
```diff
@@ -453,6 +456,17 @@ export interface SearchStatusState {
     isLoading?: boolean;
 }
 
+// Warning: (ae-forgotten-export) The symbol "EnumOrLiteral" needs to be exported by the entry point index.d.ts
```
nmanu1, line 459: Do we want to export this?

src/answers-headless.ts
cea2aj: Since UniversalAutocomplete has no impact on the state, there's not really a reason to prevent it from being ran on a vertical. The autocomplete functions are more stateless than regular universal and vertical search. Autocomplete for vertical and filter search require a verticalKey so an error there is required

etc/answers-headless.api.md
cea2aj: yep! updated

src/models/index.ts
```diff
@@ -17,6 +17,7 @@ export { VerticalSearchState, AllResultsForVertical } from './slices/vertical';
 export { BoundedRange, RangeBoundary } from './utils/boundedrange';
 export { SelectableFilter } from './utils/selectablefilter';
 export { SearchTypeEnum, SearchType } from './utils/searchType';
+export { EnumOrLiteral } from '../utils/types';
```
nmanu1, line 20: Should this export be in src/index.ts instead? I thought src/models/index.ts should only export things that are within the `models` folder

cea2aj: good point, I moved it

nmanu1 (approved):

PR #72, opened 2022-01-10
Update searchThroughFacet

nmanu1:
Changes `searchThroughFacet` to be case insensitive (for both substring matching and Levenshtein matching).

J=SLAP-1782
TEST=auto, manual

Added additional tests to verify case-insensitivity and that search terms with length less than 3 characters return correct facet options without using Levenshtein. See this behavior reflected in the sample app.

### DIFF
--- a/src/answers-utilities.ts
+++ b/src/answers-utilities.ts
@@ -4,7 +4,7 @@ import { isLevenshteinMatch } from './utils/searchable-facets';
 export default {
   /**
    * Searches through the specified facet and filters out the options that aren't a
-   * close match for the given searchTerm.
+   * close match for the given searchTerm. The comparison is case insensitive.
    *
    * @param facet - The facet whose options are searched through
    * @param searchTerm - The search term to compare the facet options against
@@ -14,7 +14,9 @@ export default {
   searchThroughFacet(facet: DisplayableFacet, searchTerm: string): DisplayableFacet {
     return {
       ...facet,
-      options: facet.options.filter(o => isLevenshteinMatch(o.displayName, searchTerm))
+      options: facet.options.filter(o =>
+        isLevenshteinMatch(o.displayName.toLowerCase(), searchTerm.toLowerCase())
+      )
     };
   }
 };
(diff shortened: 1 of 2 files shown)


oshi97 (approved):

tmeyer2115 (commented):
What about the other behavior referenced in the item? Specifically, Rose and I were seeing that no options would appear unless 3 characters were in the `input`. Is that due to the Levenshtein distance metric?

nmanu1:
> What about the other behavior referenced in the item? Specifically, Rose and I were seeing that no options would appear unless 3 characters were in the `input`. Is that due to the Levenshtein distance metric?

Levenshtein is only used if the search input is 3 characters or longer. But before this commit, I was seeing options appearing with less than 3 characters as long as the substrings matched exactly (case-sensitive). Is that not what you and Rose are seeing?

tmeyer2115:
> > What about the other behavior referenced in the item? Specifically, Rose and I were seeing that no options would appear unless 3 characters were in the `input`. Is that due to the Levenshtein distance metric?
> 
> Levenshtein is only used if the search input is 3 characters or longer. But before this commit, I was seeing options appearing with less than 3 characters as long as the substrings matched exactly (case-sensitive). Is that not what you and Rose are seeing?

We saw behavior somewhat similar to that. In the example we saw, American Express was one of the Facet options. If you typed 'a', you saw the option, but with 'am' it disappeared. Could we make the sub-string match be case-insensitive as well? I don't think that would fix the quirk Rose and I saw, but it should be an improvement. I will get the Experience from her where we saw this wonkiness.

nmanu1:
> We saw behavior somewhat similar to that. In the example we saw, American Express was one of the Facet options. If you typed 'a', you saw the option, but with 'am' it disappeared. Could we make the sub-string match be case-insensitive as well? I don't think that would fix the quirk Rose and I saw, but it should be an improvement. I will get the Experience from her where we saw this wonkiness.

The reason 'a' matched American Express before was because it was matching the second 'a' in American. But, 'am' would no longer match any substring. The change made here makes both the substring matching and the Levenshtein matching case insensitive, so that behavior should now be fixed.

https://user-images.githubusercontent.com/88398086/148955801-e36f68ee-41f1-45ab-aa96-90edb62fc68a.mov

tmeyer2115 (approved):

PR #73, opened 2022-01-11
Update version to v1.0.0.

tmeyer2115:


yen-tt:
probably need to update answers-core version to non-beta as well?

tmeyer2115:
> probably need to update answers-core version to non-beta as well?

Good catch! Done!

nmanu1 (approved):

PR #76, opened 2022-01-12
Add text matching utility function

nmanu1:
Adds the `isCloseMatch` function in answers-utilities to perform case-insensitive Levenshtein matching. This is used in `searchThroughFacet`.

J=SLAP-1802
TEST=auto

See that jest tests for `searchThroughFacet` pass.


src/answers-utilities.ts
```diff
@@ -14,9 +14,20 @@ export default {
...
-      options: facet.options.filter(o =>
-        isLevenshteinMatch(o.displayName.toLowerCase(), searchTerm.toLowerCase())
-      )
+      options: facet.options.filter(o => isCloseMatch(o.displayName, searchTerm))
     };
-  }
-};
\ No newline at end of file
+  },
+  isCloseMatch
```
tmeyer2115, line 20: Does this export work properly? There can only be one `default` export per module. It looks as if we're attempting to have two default exports. If this module is to export both `searchThroughFacet` and `isCloseMatch`, one should either be the `default` or we should have two named exports.

oshi97: this is exporting an single object that has two functions. I think I would just prefer 2 named exports, and an `import * as AnswersUtilities from './answers-utilities` wherever this gets imported if we really want to group things

nmanu1: It's not giving any errors with the export like this and I'm able to use it in the starter repo locally. But, I can switch to having two named exports instead. That would also remove the forgotten exports warning in the API report.

tmeyer2115:
Remember to update the target branch here. We can no longer merge directly into `main`.

oshi97 (approved):

PR #78, opened 2022-01-24
Add query rules actions data

nmanu1:
Add `QueryRulesActionsData` to the state.

J=SLAP-1829
TEST=auto

Check that the added jest tests pass.


tests/integration/queryrules.ts
```diff
@@ -0,0 +1,63 @@
...
+describe('AnswersHeadless queryRules interactions properly update state', () => {
+  it('executeVerticalQuery properly updates queryRules state', async () => {
+    const answers = createMockedAnswersHeadless({
+      verticalSearch: mockSearchWithQueryRules
+    }, initialState);
+    answers.setVertical('123');
+    await answers.executeVerticalQuery();
+    const expectedState = {
+      queryRules: {
+        ...initialState.queryRules,
```
cea2aj, line 40: Is this `...initialState.queryRules` line here and in the other test doing anything? It looks like it's just getting overridden by the actions below it

cea2aj:
I saw in the spec that there was some discussion about whether or not QueryRules should go in the state. Did anything come of that discussion?

nmanu1:
> I saw in the spec that there was some discussion about whether or not QueryRules should go in the state. Did anything come of that discussion?

Yes, I looked into how it would work if the query rules data were not stored in state and it would have been tricky because the response would be returned from where the search is fired (e.g. the SearchBar component). And getting that data to where it would be used (e.g. the page with the results) would have required the data to be passed around between components. I talked with Tom and we thought it would be simpler to just add it to the state.

tests/integration/queryrules.ts
nmanu1: Yeah, that was leftover from something else I was trying. I'll remove it!

cea2aj:
Look good! I think we'll just want to hold off on merging into develop until we get the new core version

oshi97 (approved):

PR #81, opened 2022-02-23
Update setVertical and setUniversal

nmanu1:
Reset the existing filters, results, direct answer, query rules custom data, and search status to their initial states when switching to a new searcher using `setUniversal` or `setVertical` because these states pertain only to the previous searcher.

J=SLAP-1910
TEST=auto

Check that the updated jest tests pass.

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -27,6 +27,12 @@ import * as answersUtilities from './answers-utilities';
 import { SelectableFilter } from './models/utils/selectablefilter';
 import { transformFiltersToCoreFormat } from './utils/transform-filters';
 import { SearchTypeEnum } from './models/utils/searchType';
+import { initialState as initialVerticalState } from './slices/vertical';
+import { initialState as initialUniversalState } from './slices/universal';
+import { initialState as initialFiltersState } from './slices/filters';
+import { initialState as initialDirectAnswerState } from './slices/directanswer';
+import { initialState as initialQueryRulesState } from './slices/queryrules';
+import { initialState as initialSearchStatusState } from './slices/searchstatus';
 
 /**
  * Provides the functionality for interacting with an Answers Search experience.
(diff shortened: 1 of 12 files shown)


src/answers-headless.ts
```diff
@@ -81,6 +81,12 @@ export default class AnswersHeadless {
    * @param verticalKey - The vertical key to set
    */
   setVertical(verticalKey: string): void {
+    this.setState({
+      ...this.state,
+      filters: {},
```
cea2aj, line 86: Are there other things in state that we might want to clear? For example, it might make sense to clear DirectAnswer as well

tests/integration/query.ts
```diff
@@ -112,7 +112,7 @@ describe('ensure correct results from latest request', () => {
   });
 
   it('vertical search get correct results based on up-to-date response', async () => {
-    answers.addListener({
+    const removeListener = answers.addListener({
```
oshi97, line 115: would it be a little simpler to just make a new instance of `answers` in each test? otherwise I think it's possible for people to add additional tests and forget to cleanup the listener.

I would probably just pull all the initialization into a function and call it at the start of each test. you could also use the whole pattern of re-initng it in the beforeEach, though I personally find that way a little more confusing (and I guess it's less "functional" with regards to functional programming)

```ts
let answers;
beforeEach(() => {
  answers = blah
})
```

src/answers-headless.ts
nmanu1: Yeah, I updated to also clear direct answer, query rules, and search status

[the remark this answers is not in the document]
nmanu1: makes sense, updated!

tests/integration/query.ts
nmanu1: updated to the first approach so there's a function that's called at the start of each test to make a new instance of `answers`

tests/integration/query.ts
```diff
@@ -134,11 +118,13 @@ describe('ensure correct results from latest request', () => {
     expect(answers.state.query.input).toEqual(queries[2]);
     expect(answers.state.vertical.results).toEqual([queries[2]]);
     expect(updateResult.mock.calls).toHaveLength(2);
+    removeListener();
```
oshi97, line 121: I think we shouldn't need these remove listener calls anymore?

oshi97: I mean, we don't remove listeners when somebody closes their webpage right? I feel like having code that isn't needed is confusing to people who look at the tests in the future and have to wonder why this listener is being removed. It might lead to confusion like "Are they removing the listener because the headless instance is being shared?".

nmanu1: That makes sense, updated!

oshi97 (approved):

oshi97 (approved):
lgtm!

PR #82, opened 2022-02-28
Store display name for static filters in state

yen-tt:
update FilterState.static to use DisplayableFilter interface instead of SelectableFilter, which include a `displayName` field to store the value use to display on page.

J=SLAP-1906
TEST=manual&auto

smoke test with answers PB repo's page with facets and filterSearch components.
passed jest tests


src/models/utils/displayableFilter.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export interface DisplayableFilter extends Filter {
+  /**
+   * Whether or not the filter is selected.
+   */
+  selected?: boolean,
+  /**
+   * The filter's display name.
+   */
+  displayName?: string
```
oshi97, line 16: should displayName and selected be optional?

src/slices/filters.ts
```diff
@@ -63,15 +64,15 @@ const reducers = {
     if (!state.static) {
       state.static = [];
     }
-    const { filter: targetFilter, shouldSelect } = payload;
-    const matchingFilter = state.static.find(storedSelectableFilter => {
-      const { selected:_, ...storedFilter } = storedSelectableFilter;
+    const { filter: targetFilter, shouldSelect, displayName } = payload;
+    const matchingFilter = state.static.find(storedDisplayableFilter => {
+      const { selected:_, displayName:__, ...storedFilter } = storedDisplayableFilter;
```
oshi97, line 69: if we don't remove the unused properties (`selected` and `displayName`) does typescript complain? was thinking it might be slightly nicer to just delete line 69

yen-tt: It doesn't complain, will update!

src/slices/filters.ts
```diff
@@ -14,13 +14,14 @@ interface FacetPayload {
 
 interface FilterPayload {
   filter: Filter
```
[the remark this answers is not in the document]
oshi97: Ah yeah I think we might have had `selected` be separate from the passed in filter at one point in `AnswersHeadless.setFilterOption` (or whatever we had at the time)

src/models/utils/displayableFilter.ts
yen-tt: based on previous discussions where headless may have use cases without UI involved, I thought making displayName optional might be more flexible. But yeah selected should be required

oshi97 (approved):

cea2aj:
Hey Yen, this is a breaking change since any code which imports `SelectableFilter` will break, and any code which calls `setFilterOption` or s`setStaticFilters` will also break. Therefore, this change would warrant a version bump from major version 1 to 2, which I don't think we are ready for yet. We'd want to confirm with @tmeyer2115.. Most likely, I expect he'll want us to make this change backwards compatible. We can do that by not removing `SelectableFilter` and updating the code to use either `SelectableFilter` or `DisplayableFilter`. `SelectableFilter` can be marked as deprecated.

PR #83, opened 2022-03-01
keep SelectableFilter for backward compatibility

yen-tt:
breaking changes in [another pr](https://github.com/yext/answers-headless/pull/82) when adding DisplayableFilter will require a major version bump. For backward compatibility support in current major version, this pr add displayName as an optional additional field to SelectableFilter instead of having new interface DisplayableFilter

J=SLAP-1906
TEST=manual&auto

passed jest tests
test with answers PB repo's page. See that answers-react-component recognize SelectableFilter in state.filters.static and build successfully.


etc/answers-headless.api.md
```diff
@@ -498,6 +498,11 @@ export enum SearchTypeEnum {
     Vertical = "vertical"
 }
 
+// @public @deprecated
+export interface SelectableFilter extends Filter {
+    selected: boolean;
```
cea2aj, line 503: After discussing this approach with Tom, we came to the conclusion that it would be simpler to add an optional `displayName` to `SelectableFilter` rather than deprecating `SelectableFilter` and creating `DisplayableFilter`. Strictly speaking, adding an optional property is breaking change, however it is less much less likely to cause issues compared to removing `SelectableFilter` altogether.

yen-tt: I see. This approach wouldn't have consistent naming like what we have with `DisplayableFacet` from core. If we are introducing some type of breaking change either way, I think we should rename altogether in the next major version. But yea this would make it complicated for backward compatibility so adjusting SelectableFilter to have displayName make sense. I will update :+1:

cea2aj (approved):
🚀

PR #85, opened 2022-03-04
handle core rejected promise in universal/vertical search

yen-tt:
update `executeUniversalQuery` and `executeVerticalQuery` to handle rejected promise/error throw from core. Using try/catch, if there's a rejected promise from latest request, ensures isLoading state is set back to false and data in httpManager is up to date, before returning a rejected promise.

J=SLAP-1941
test=auto

see new jest tests fail without the changes. See all jest tests pass with changes from pr.

### DIFF
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -330,7 +330,7 @@ export default class AnswersHeadless {
       if (isLatestResponse) {
         this.stateManager.dispatchEvent('searchStatus/setIsLoading', false);
       }
-      throw e;
+      return Promise.reject(e);
     }
     const isLatestResponse = this.httpManager.processRequestId('universalQuery', thisRequestId);
     if (!isLatestResponse) {
--- a/src/answers-headless.ts
+++ b/src/answers-headless.ts
@@ -308,7 +308,7 @@ export default class AnswersHeadless {
     const { referrerPageUrl, context } = this.state.meta;
     const { userLocation } = this.state.location;
 
-    const response = await this.core.universalSearch({
+    const request = {
       query: input || '',
       querySource,
       queryTrigger,
@@ -320,7 +320,19 @@ export default class AnswersHeadless {
       context,
       referrerPageUrl,
       restrictVerticals
-    });
+    };
+
+    let response = undefined;
+    try {
+      response = await this.core.universalSearch(request);
+    } catch (e) {
+      const latestResponseId = this.httpManager.getLatestResponseId('universalQuery');
+      if (thisRequestId > latestResponseId) {
+        this.stateManager.dispatchEvent('searchStatus/setIsLoading', false);
+        this.httpManager.setResponseId('universalQuery', thisRequestId);
+      }
+      return Promise.reject(e);
+    }
 
     const latestResponseId = this.httpManager.getLatestResponseId('universalQuery');
     if (thisRequestId < latestResponseId) {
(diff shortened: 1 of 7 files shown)


src/answers-headless.ts
```diff
@@ -320,7 +320,19 @@ export default class AnswersHeadless {
...
       referrerPageUrl,
       restrictVerticals
-    });
+    };
+
+    let response: UniversalSearchResponse;
+    try {
+      response = await this.core.universalSearch(request);
+    } catch (e) {
+      const latestResponseId = this.httpManager.getLatestResponseId('universalQuery');
```
cea2aj, line 329: what do you think of adding a helper to httpManager to make this logic a little simpler? For example, we could have `isMostRecentUniversalRequest` which takes in a requestId and returns a boolean. I'm also considering having a `processUniversalRequestId` which returns whether or not it's the most recent id as a boolean, and if so, it automatically sets the responseId. I think either of those might help make this logic a tad more clear and concise. Such a change could be used elsewhere in headless so I'm okay if we make another item and consider it out of scope of this PR

yen-tt: sure! I think the `processUniversalRequestId` approach sounds good, but probably make it a generic `processRequestId` that takes in a request type (e.g. 'universalQuery') as well. I can update

src/answers-headless.ts
```diff
@@ -320,7 +320,19 @@ export default class AnswersHeadless {
...
+      if (thisRequestId > latestResponseId) {
+        this.stateManager.dispatchEvent('searchStatus/setIsLoading', false);
+        this.httpManager.setResponseId('universalQuery', thisRequestId);
+      }
+      return Promise.reject(e);
```
oshi97, line 334: just out of curiosity does `throw e` here have the same behavior as returning a Promise.reject(e)?

oshi97: Also what information does this error have? like do we need to add any extra info onto it?

yen-tt: I believe so, both would stop the remaining code from running, and requires some rejected promise handling logic from the caller (it will receive a rejected promise in both case)

yen-tt: I think the error message coming from core should be sufficient, unless you have something else in mind we should add?

oshi97: Ah I just wasn't sure what the message was. I was mostly wondering whether it said what kind of request failed. The API doesn't always have the best error messages from what I remember

oshi97: Though if the error msg doesn't have that info maybe that's core's job

src/http-manager.ts
```diff
@@ -38,4 +38,22 @@ export default class HttpManager {
   getLatestResponseId(responseName: ServiceType): number {
     return this.latestResponseIds[responseName];
   }
+
+  /**
+   * Update the recorded received response id of the given service type if
+   * the given request id corresponds to the latest response.
```
cea2aj, line 44: what do you think of rewording to 'the given request id is newer than the latest saved response id?'

src/answers-headless.ts
yen-tt: just tested with a failed response from filterSearch, although you can look at the stack trace and network, it looks like the error msg itself doesn't know what kind of request it is. I think this is something we can update in core instead tho

oshi97: I'll make an item!

cea2aj (approved):
lgtm!

PR #86, opened 2022-03-16
Answers Agent Support

nmanu1:
With each request, pass the additional HTTP headers. Currently, this only allows the Client-SDK header, which includes the ANSWERS_HEADLESS agent specifying the Headless version used by the site, as well as any additional agents passed to `provideAnswersHeadless`. Add unit tests to confirm this behavior.

The `provideAnswersHeadless` function is overloaded so we can pass the headers when using the internal signature but not expose the param publicly.

Like in Core, because we are importing package.json to get the version, in order to get the `commonjs` and `esm` imports to work correctly, the `outDir` was changed from `lib` to `dist`.

J=SLAP-1950
TEST=auto

Check that Jest tests pass.


src/index.ts
```diff
@@ -62,8 +69,9 @@ export function provideAnswersHeadless(config: HeadlessConfig): AnswersHeadless
   const stateManager = new ReduxStateManager(
     store, headlessId || DEFAULT_HEADLESS_ID, headlessReducerManager);
   const httpManager = new HttpManager();
+  const customClientSdk = getAdditionalHttpHeaders(additionalHttpHeaders);
```
oshi97, line 72: can we rename customClientSdk?

src/models/client-sdk.ts
```diff
@@ -0,0 +1,26 @@
+import { AdditionalHttpHeaders, ClientSDKHeaderValues } from '@yext/answers-core';
+
+/**
+ * Allows users to specify additional values for specific HTTP headers.
+ *
+ * @public
+ */
+export interface CustomHttpHeaders extends AdditionalHttpHeaders {
```
oshi97, line 8: can we alias the original AdditionalHttpHeaders as something else like `import { AdditionalHttpHeaders as blah }` and then export a new interface called AdditionalHttpHeaders? I'm not sure if the export-star eslint plugin yells at us for that but it could be hard to come up with a new name every time we add a new library to this "additional http header stack" with every new lib we add on top

oshi97: though, we have answers-core-re-exports so maybe it won't complain after all?

oshi97: since we can just remove it from there

nmanu1: I thought about doing that first, but the problem is that we're exporting everything from Core in Headless, so when API extractor builds the docs for those, it would resolve any inheritDoc references to `AdditionalHttpHeaders` as the new one we specify in Headless. we can rename each place we import it from Core (including answers-core-re-exports) but we wouldn't be able to change references in the docs

cea2aj: @nmanu1, I think that behavior sounds okay to me? If you're looking at the docs for `AdditionalHttpHeaders` in the Headless Docs, I would expect to see the model which includes `ANSWERS_HEADLESS`.

nmanu1: Ok, in that case, I think most of the time it would be fine. The only place where it might be odd is if we use Core's `AdditionalHttpHeaders` as a type somewhere in Headless (such as in the AnswersHeadless constructor). Even though we rename it something else when we import it, I think the docs still point to `AdditionalHttpHeaders`, which would now resolve to the interface defined in Headless. That would be slightly inaccurate for the linked type, but if we're ok with that, I can make the change

cea2aj: Why are we using Core's `AdditionalHttpHeaders` in the `AnswersHeadless` constructor?

nmanu1: The AnswersHeadless constructor stores the exact headers that would be sent to Core, which includes the Headless agent. This would not be allowed if we used the interface defined in Headless as the type (which doesn't allow the Headless agent to be specified)

cea2aj: In that case, maybe it would be simpler to stick with the current approach and perhaps consider some renaming? `CustomHttpHeaders` could become `HeadlessAdditionalHttpHeaders` and  `CustomClientSDKHeaderValues` could become `ClientSDKHeaderValuesExcludingHeadless`

etc/answers-headless.api.md
```diff
@@ -168,6 +184,16 @@ export function createNearMeFilter(position: NearFilterValue): Filter;
 // @public
 export function createNumberRangeFilter(fieldId: string, range: BoundedRange<number>): FilterTypes;
 
+// @public
+export interface CustomClientSDKHeaderValues extends ClientSDKHeaderValues {
+    'ANSWERS_HEADLESS'?: never;
+}
+
+// @public
+export interface CustomHttpHeaders extends AdditionalHttpHeaders {
```
cea2aj, line 193: Does this need to extend AdditonalHttpHeaders? It looks like it's overwriting the 'Client-SDK' prop anyway. so I'm not sure what benefit the `extends` gives us

nmanu1: At the moment, it isn't necessary. But, I thought it might be better to extend it so that in the future, if more headers are allowed, we don't need to change this to allow them.

tests/unit/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+jest.mock('../../src/answers-headless.ts');
+
+const baseConfig = {
+  apiKey: 'api-key',
+  experienceKey: 'exp-key',
+  locale: 'en'
+};
+
+const headers = getAdditionalHttpHeaders();
```
cea2aj, line 12: nit: what do you think of renaming this to `headlessHeaders` since the only purpose of this is getting the header containing `ANSWERS_HEADLESS`

src/models/client-sdk.ts
nmanu1: Sure, I can rename. That would give us a consistent way to name this in other libraries if we build on top of this

cea2aj: In my experience, renaming can cause confusion about which interface is actually being used. I prefer additional interfaces with slightly longer names if it increases clarity and provides us with accurate documentation. I also considered removing the static type checking to ensure ANSWERS_HEADLESS isn't supplied and simply relying on runtime checks, however in the past we have decided to use more complicated types for the sake of type checking at build time (e.g. supporting apiKey or auth token but not both), and I think we should make that same decision here as well.

src/index.ts
```diff
@@ -62,8 +69,9 @@ export function provideAnswersHeadless(config: HeadlessConfig): AnswersHeadless
   const stateManager = new ReduxStateManager(
     store, headlessId || DEFAULT_HEADLESS_ID, headlessReducerManager);
   const httpManager = new HttpManager();
+  const httpHeaders = getAdditionalHttpHeaders(additionalHttpHeaders);
```
oshi97, line 72: would it be bad if we did this inside the AnswersHeadless constructor instead of in provideAnswersHeadless? It makes things slightly less flexible, but also lets us cut down on the number of interfaces we're exporting for this feature? We won't need to export both HeadlessAdditionalHttpHeaders, and ALSO AdditionalHttpHeaders. I think that might result in an overall nicer developer experience. It'll also match Rose's spec more closely

nmanu1: do you mean have `provideAnswersHeadless` pass `additionalHttpHeaders` directly to the `AnswersHeadless` constructor and have the constructor call the `getAdditionalHttpHeaders` method?
I'm not sure how that would cut down on the imports? we would still need to export both `HeadlessAdditionalHttpHeaders` (as part of `HeadlessConfig`) and `AdditionalHttpHeaders` because it's extended by `HeadlessAdditionalHttpHeaders` and referenced in Core docs

oshi97: OH I see I thought we wouldn't need to export AdditionalHttpHeaders if we were only extending it, but api-extractor means we have to

oshi97: could you check with Rose on the interface naming then? I think there's a chance she'll have some thoughts on shortening some of them. I was thinking if we could export less then we could simplify the names some, but it looks like that's not possible

oshi97: specifically I feel like `ClientSDKHeaderValuesExcludingHeadless` is a little wordy? at the very least maybe something like `HeadlessClientSdkHeaderValues`?

nmanu1: sure, I'll ask her

src/models/client-sdk.ts
```diff
@@ -0,0 +1,26 @@
+import { AdditionalHttpHeaders, ClientSDKHeaderValues } from '@yext/answers-core';
+
+/**
+ * Allows users to specify additional values for specific HTTP headers.
+ *
+ * @public
+ */
+export interface HeadlessAdditionalHttpHeaders extends AdditionalHttpHeaders {
```
tmeyer2115, line 8: Do we really need this additional `interface` here? I'm not sure how much value it really adds and it adds complexity to the code and mental model. Could we not just have the `interface` below and then do something like:
```
const values: ClientSDKHeaderValuesExcludingHeadless = blah;
const additionalHeaders: AdditionalHttpHeaders = {
     'Client-SDK': values
};
```

src/index.ts
```diff
@@ -32,7 +34,11 @@ export type HeadlessConfig = AnswersConfig & {
    * The verticalKey associated with the vertical to manage. If none is provided,
    * Answers Headless will manage universal search.
    */
-  verticalKey?: string
+  verticalKey?: string,
+  /**
+   * {@inheritDoc HeadlessAdditionalHttpHeaders}
+   */
+  additionalHttpHeaders?: HeadlessAdditionalHttpHeaders
```
tmeyer2115, line 41: @cea2aj, Is there anyway we can make this property hidden? I don't think there's much value in exposing it to users of Headless at the moment. Headless operates at a higher level of abstraction than Core, so it's reasonable to assume they won't need to know anything about this. Only our libraries will leverage this property internally. 

Depending on what other Headers we make available in the future, we could make this public. But, to start with, I think having it be hidden in the docs could be good. I don't care that it's still visible in the Constructor. If you're building Headless yourself (not through `provideCore`) you're a power user anyway.

src/models/client-sdk.ts
nmanu1: I added this interface to mimic the structure in Core. `AdditionalHttpHeaders` was added in Core because we might allow more headers in the future. In that case, I assumed we would allow more headers to be specified in Headless as well (through `provideAnswersHeadless`).
If we didn't have the `HeadlessAdditionalHttpHeaders` interface, would we just accept an object of type `ClientSDKHeaderValuesExcludingHeadless` in `provideAnswersHeadless` and populate `additionalHeaders` in the function?

src/index.ts
cea2aj: Unfortunately because `HeadlessConfig` is a type and not an interface, we can't make this particular property internal, however since `HeadlessAdditionalHttpHeaders` is an interface so we could make its own properties internal. When when mark a property internal, it's excluded from the docs and the type bundle.

cea2aj: Currently API Extractor isn't trimming the type bundle, so if we do add an `@internal` tag, we'd need to disable the `untrimmedFilePath` and enable the `publicTrimmedFilePath`

cea2aj: Alternatively, we could remove `additionalHttpHeaders` from `HeadlessConfig` altogether and we supply the Headless Agent info behind the scenes. That would let us avoid making these additional interfaces.

oshi97 (approved):
if this works with headless-react lgtm!

nmanu1:
> if this works with headless-react lgtm! we might want to do a quick manual test?

I hooked it up locally to headless-react and the starter app and it works!

src/index.ts
```diff
@@ -48,7 +49,30 @@ const headlessReducerManager = new HeadlessReducerManager();
  *
  * @public
  */
-export function provideAnswersHeadless(config: HeadlessConfig): AnswersHeadless {
+export function provideAnswersHeadless(config: HeadlessConfig): AnswersHeadless;
+
+/**
+ * Supplies a new instance of {@link AnswersHeadless}, using the provided configuration,
```
cea2aj, line 55: nit: I was thinking we could put the docs above the just above the public implementation to avoid duplication

cea2aj (approved):

PR #88, opened 2022-03-23
Add Filter and Facet NumberRangeValue support

cea2aj:
Add support for facets and filters with a NumberRangeValue


oshi97 (approved):

tests/unit/slices/filters.ts
```diff
@@ -101,6 +101,44 @@ describe('filter slice reducer works as expected', () => {
     expect(actualState).toEqual(selectedExpectedState);
   });
 
+  it('setFilterOption action is handled properly with a Filter with a NumberRangeValue', () => {
+    const selectedFilter: SelectableFilter = {
+      fieldId: 'id1',
+      matcher: Matcher.Equals,
```
nmanu1, line 107: does this need to be `Matcher.Between` for a number range?

tests/unit/slices/filters.ts
```diff
@@ -101,6 +101,44 @@ describe('filter slice reducer works as expected', () => {
     expect(actualState).toEqual(selectedExpectedState);
   });
 
+  it('setFilterOption action is handled properly with a Filter with a NumberRangeValue', () => {
+    const selectedFilter: SelectableFilter = {
```
nmanu1, line 105: can some of these filter fields be defined at the start of the test suite so they can be used by both number range tests? it seems like they're mostly the same in both

tests/unit/slices/filters.ts
cea2aj: sure, I updated to `Matcher.Between`

nmanu1 (approved):

PR #89, opened 2022-03-24
use shared eslint

oshi97:
Fixed only auto-fixable issues. Added the language-subtag-registry preamble and distributed third party notices.

J=SLAP-2003, SLAP-2064
TEST=build


cea2aj (approved):

cea2aj (changes requested):
oh, looks like snyk is failing

package.json
```diff
@@ -9,7 +9,8 @@
   "files": [
     "dist",
     "lib",
-    "src"
+    "src",
+    "THIRD-PARTY-NOTICES"
```
oshi97, line 13: Oh omg this was a nice catch. Would we need to add this to other repos? (maybe a separate item if we do?)

cea2aj: Ya, we should add this to everything we distribute. AFAIK, we only have it in answers-react-components and in slapshot-tools/packages/eslint-config-slapshot

cea2aj (approved):

package.json
oshi97: Do we also need it for the LICENSE file?

oshi97: Ah nvm LICENSE is automatically included

oshi97:
I can't +2 since I opened the original PR but lgtm!

PR #95, opened 2022-05-18
redux changes to support react 18

yen-tt:
Using react 18 with the current answers-headless would result in peer dependency incompatibility issues with redux. Upgrade redux-toolkit to 1.8.1 ([release notes](https://github.com/reduxjs/redux-toolkit/releases/tag/v1.8.1)). [Support react version](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/package.json#L108) at this time:  `"react": "^16.9.0 || ^17.0.0 || ^18"`
Screen Shot 2022-05-18 at 3 51 45 PM
This upgrade will result in an error from store.dispatch call, specifically some of redux-toolkit's 1.8 changes to middleware dispatch types may lead to the dispatch function to be typed as never in TypeScript.
Screen Shot 2022-05-18 at 3 54 53 PM
This can be resolve by specifying the third generic type for Middleware in EnhancedStore interface, which is:
`export interface EnhancedStore<S = any, A extends Action = AnyAction, M extends Middlewares<S> = Middlewares<S>> extends Store<S, A>`

From [documentation](https://github.com/reduxjs/redux-thunk#redux-toolkit) on redux-thunk github page, since we are using redux-toolkit already, there's no need to install redux-thunk separately so that's removed from package.json's deps.

Note: will release an alpha version containing this change to use in answer-headless-react's alpha version as well.

J=SLAP-2095
TEST=manual

see that `npm run build` was successful. When used locally with answers-headless-react and answers-react-components, hooked to the test-site with React 18, the pages work as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless",
-  "version": "1.1.0-beta.12",
+  "version": "1.1.1-alpha.0",
   "description": "",
   "author": "slapshot@yext.com",
   "license": "ISC",
@@ -22,11 +22,10 @@
     "generate-notices": "generate-license-file --input package.json --output THIRD-PARTY-NOTICES --overwrite"
   },
   "dependencies": {
-    "@reduxjs/toolkit": "^1.7.0",
+    "@reduxjs/toolkit": "^1.8.1",
     "@yext/answers-core": "^1.6.0-beta.7",
     "js-levenshtein": "^1.1.6",
-    "lodash": "^4.17.21",
-    "redux-thunk": "^2.4.1"
+    "lodash": "^4.17.21"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.16.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless",
-  "version": "1.1.1-alpha.0",
+  "version": "1.1.1-alpha.0-95",
   "description": "",
   "author": "slapshot@yext.com",
   "license": "ISC",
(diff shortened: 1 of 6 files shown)


oshi97:
Have we done any testing with the new concurrent features? I feel like we may need to use `useSyncExternalStore` (just based on reading through https://reactjs.org/blog/2022/03/29/react-v18.html#usesyncexternalstore) somewhere, probably in answers-headless-react.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless",
-  "version": "1.1.0-beta.12",
+  "version": "1.1.1-alpha.0",
```
nmanu1, line 3: I think we decided that alpha/beta tags should include the commit date in the name?

yen-tt: Will update after the pr is ready to publish (although now I'm thinking if we want a commit hash instead of a date instead)

yen-tt:
> Have we done any testing with the new concurrent features? I feel like we may need to use `useSyncExternalStore` (just based on reading through https://reactjs.org/blog/2022/03/29/react-v18.html#usesyncexternalstore) somewhere, probably in answers-headless-react. I think we probably won't need changes to answers-headless but I would like to test first.

Did some testing with the concurrent features, we would need to make changes in answers-headless-react but answers-headless can remain as is. If we do implement this, which is a react 18 feature, I'm thinking we may use something like `use-sync-external-store` [npm](https://www.npmjs.com/package/use-sync-external-store) package which supports ^16.8.0 || ^17.0.0 || ^18.0.0".

oshi97 (approved):

package.json
nmanu1: Would the PR number be better since it's known before merge and gives more granularity than commit date?

yen-tt: I'm thinking potentially a github action that checks a pr commit message after merging to develop for a specific string (maybe `PUBLISH='1.1.1-alpha.0'`) where it check in develop branch, update the package version locally, and publish an alpha version, maybe make an alpha branch too.

PR #96, opened 2022-05-26
export SandboxEndpoints from answers-core + other new exports

oshi97:
This PR was made for the purpose of exporting the new SandboxEndpoints
constant from answers-core.

This new version of answers-core has a few other references types that
need to be exported, otherwise they show up as undocumented in the api-documentor
docs.

J=SLAP-2106
TEST=manual

check that documentation has no missing exports
put in sandbox creds to node test site, see a search run


yen-tt (approved):

package.json
```diff
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@reduxjs/toolkit": "^1.8.1",
-    "@yext/answers-core": "^1.6.0-beta.7",
+    "@yext/answers-core": "^1.7.0-alpha.165",
```
nmanu1, line 26: do we also need to update the headless version to publish a new version? I wasn't sure if we were trying to export these endpoints from headless-react or the component library too

oshi97: I forgot to bump the version! updated now

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-headless",
-  "version": "1.1.1-alpha.0-95",
+  "version": "1.1.1-alpha.0-96",
```
nmanu1, line 3: I think we can remove the `0-` and just have `1.1.1-alpha.95` to match our other libraries

oshi97: updated!

PR #104, opened 2022-06-21
edited license to package.json and updated year in LICENSE

alextaing:
Edited license to BSD-3-Clause in package.json and package-lock.json, and updated year to 2022 in LICENSE.

J=SLAP-2161
TEST=none


oshi97:
could you try running npm install to double check that it doesn't update the package-lock.json? (this is something we miss pretty commonly when editing the package.json)

oshi97 (approved):

PR #108, opened 2022-06-29
add descripiton and keywords to package.json

juliannzhou:
This PR adds description and keywords to package.json. The description is copied from README and keywords provided by product.

J=SLAP-2190
TEST=manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,20 @@
 {
   "name": "@yext/answers-headless",
   "version": "1.2.0",
-  "description": "",
+  "description": "A library for powering UI components for Yext Answers integrations",
   "author": "slapshot@yext.com",
   "license": "ISC",
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
+  "keywords": [
+    "search",
+    "nlp",
+    "ai",
+    "answers",
+    "yext",
+    "react",
+    "components"
+  ], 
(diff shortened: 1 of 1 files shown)


oshi97 (approved):

package.json
```diff
@@ -1,11 +1,20 @@
...
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
+  "keywords": [
+    "search",
+    "nlp",
+    "ai",
+    "answers",
+    "yext",
+    "react",
+    "components"
```
oshi97, line 16: Did product ask for "components" and "react" here? This library doesn't have any components in it, and also doesn't use react

oshi97: I would probably put "redux" here if anything

nmanu1 (commented):
can this PR be merged into `develop` instead of `main`

package.json
juliannzhou: Product only gave me one list, but none of the repositories have keywords in package.json. I thought I'd just put the same list, but I was wondering if we need different lists of keywords for all of the repositories?

juliannzhou: I merged this branch by mistake, but I reverted the changes...

juliannzhou: I merged this into main by mistake, but I've reverted the changes.

PR #114, opened 2022-07-07
add description and keywords to package.json

juliannzhou:
This PR adds description and keywords to package.json. The description is copied from README and keywords approved by product.

J=SLAP-2190
TEST=manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,20 @@
 {
   "name": "@yext/answers-headless",
   "version": "1.2.0",
-  "description": "",
+  "description": "A library for powering UI components for Yext Answers integrations",
   "author": "slapshot@yext.com",
   "license": "ISC",
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
+  "keywords": [
+    "search",
+    "nlp",
+    "ai",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,11 +1,17 @@
 {
   "name": "@yext/answers-headless",
   "version": "1.2.0",
-  "description": "",
+  "description": "A library for powering UI components for Yext Answers integrations",
```
nmanu1, line 4: this should probably be "Search" instead of "Answers"

package.json
```diff
@@ -1,11 +1,17 @@
...
   "version": "1.2.0",
-  "description": "",
+  "description": "A library for powering UI components for Yext Search integrations",
   "author": "slapshot@yext.com",
   "license": "ISC",
   "main": "./dist/commonjs/src/index.js",
   "module": "./dist/esm/src/index.js",
+  "keywords": [
+    "Search",
+    "Yext",
```
oshi97, line 11: I would change redux and state management to lowercase. npm search is case insensitive so it probably doesn't matter but it's what I'm more used to seeing
see https://www.npmjs.com/package/redux

PR #115, opened 2022-07-11
Answers -> Search 1: package.json

ElemelonWind:
Updated package.json. Part 2 will be updating new unique customer facing components

J=SLAP-2204
TEST=auto


oshi97 (changes requested):
actually can you merge this into a feature branch instead of develop?

PR #128, opened 2022-07-26
Version 1.4.0

alextaing:
### Changes
- `matchedSubstrings` from `spellCheck` is now accessible in state (#126)
    - An array of objects with properties `offset` and `length` that represent the matching substrings between the `originalQuery` and the `correctedQuery`


alextaing:
Does the change of making `Result` a generic have an effect on how users use headless?

tmeyer2115 (approved):

nmanu1 (commented):
The version in the PR description needs to be updated

tmeyer2115 (approved):

PR #130, opened 2022-08-15
Expand static filters

nmanu1:
This PR makes the static filters state more flexible, so users can store more complex static filters as well as field value filters. It incorporates the changes in the Core models, including the new `StaticFilter` interface. The `SelectableFilter` interface is replaced with `SelectableStaticFilter` and the static filters state now accepts an array of the latter type. The Jest tests were updated to reflect these changes and to test that the new functionality works as expected.

J=SLAP-2326
TEST=auto, manual

See that the new and existing Jest tests pass. From the Node test-site, store static filters in State and run vertical searches. See that they execute successfully and give the correct results.


src/models/utils/selectableStaticFilter.ts
```diff
@@ -1,12 +1,16 @@
...
+import { StaticFilter } from '@yext/search-core';
 
 /**
- * A {@link Filter} with additional information, such as an
+ * A {@link StaticFilter} with additional information, such as an
  * optional display name and whether or not it is selected.
  *
  * @public
  */
-export interface SelectableFilter extends Filter {
```
oshi97, line 9: you don't want to extend any more?

oshi97, line 9: You don't want to extend anymore here?

nmanu1: We can't extend anymore because `StaticFilter` is a union type. TS gives this error if I try: "An interface can only extend an object type or intersection of object types with statically known members."

tests/unit/search-headless.ts
```diff
@@ -349,14 +376,26 @@ describe('search works as expected', () => {
   it('vertical search works', async () => {
     answers.state.meta.searchType = SearchTypeEnum.Vertical;
     await answers.executeVerticalQuery();
-    const { fieldId, matcher, value } = mockedState.filters.static[0];
+    const { kind, fieldId, matcher, value } = mockedState.filters.static[0].filter as FieldValueStaticFilter;
```
oshi97, line 379: are these asserts necesssary?

tests/unit/search-headless.ts
```diff
@@ -349,14 +376,26 @@ describe('search works as expected', () => {
...
       querySource: mockedState.query.querySource,
       queryTrigger: mockedState.query.queryTrigger,
       verticalKey: mockedState.vertical.verticalKey,
-      staticFilters: { fieldId, matcher, value },
+      staticFilter: {
+        kind: 'conjunction',
+        combinator: FilterCombinator.AND,
+        filters: [
+          { kind, fieldId, matcher, value },
+          { kind: disjunctionKind, combinator, filters }
```
oshi97, line 396: Could this just be `mockedState.filters.static[0]` instead of { kind: disjunctionKind, combinator, filters }?

tests/unit/slices/filters.ts
```diff
@@ -1,65 +1,130 @@
...
-    const staticFilter = [selectableFilter];
-    const actualState = reducer({}, setStatic(staticFilter));
+    const staticFilters = [selectableFilter, selectedDisjunctionFilter, selectedConjunctionFilter];
+    const actualState = reducer({}, setStatic(staticFilters));
     const expectedState = {
-      static: staticFilter
+      static: staticFilters
     };
 
     expect(actualState).toEqual(expectedState);
```
oshi97, line 117: Think this is more personal preference but what do you think of `expect(actualState).toEqual({ static: staticFilters })`?

src/utils/filter-creators.ts
```diff
@@ -89,38 +84,52 @@ function createRangeFilter(fieldId: string, range: BoundedRange<number|Date>): F
...
-export function combineFilters(
-  filterA: FilterTypes,
-  filterB: FilterTypes,
-  combinator: FilterCombinator): CombinedFilter {
+export function combineStaticFilters(
+  filterA: StaticFilter,
+  filterB: StaticFilter,
+  combinator: FilterCombinator
+): StaticFilter {
+  if (combinator === FilterCombinator.OR) {
```
oshi97, line 121: Do we need runtime checking in addition to typescript? I haven't thought super deeply about it.

src/utils/transform-filters.ts
```diff
@@ -1,60 +1,70 @@
...
-  const groupedFilters: Record<string, Filter[]> = selectedFilters.reduce((groups, element) => {
-    const { selected: _, displayName: __, ...filter } = element;
-    groups[filter.fieldId]
-      ? groups[filter.fieldId].push(filter)
-      : groups[filter.fieldId] = [filter];
-    return groups;
-  }, {});
+
+  const combinationFilters: StaticFilter[] = [];
+  const groupedFilters: Record<string, FieldValueStaticFilter[]> = selectedFilters.reduce(
```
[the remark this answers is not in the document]
oshi97: `groupedFilters` is about as vague as it gets (I'm pretty sure I came up with this variable name back in the SDK...)

src/utils/filter-utils.ts
```diff
@@ -1,16 +1,42 @@
...
+      : false;
+  }
+
+  if (otherFilter.kind === 'fieldValue') {
+    return false;
+  }
+
+  return thisFilter.combinator === otherFilter.combinator
+    && thisFilter.filters.length === otherFilter.filters.length
+    && thisFilter.filters.every((f, index) => areStaticFiltersEqual(f, otherFilter.filters[index]));
```
yen-tt, line 41: I suppose order matters? We wouldn't want to consider two filter arrays [A, B] and [B, A] be equal?

src/utils/filter-creators.ts
nmanu1: Here, we need to have some checking so that typescript doesn't complain about this case. It gives an error if you try to combine 2 arbitrary StaticFilters together because they might be conjunctions being combined with a disjunction

tests/unit/slices/filters.ts
nmanu1: it does seem a little unnecessary for simple objects like these, but I like that the `expectedState` variables are consistent with our other tests where there are more complex objects, which is why I left them in. but, if you have a strong preference, I can remove them for the simpler cases

src/utils/transform-filters.ts
```diff
@@ -1,60 +1,70 @@
...
+        fieldIdToFilters[filter.fieldId]
+          ? fieldIdToFilters[filter.fieldId].push(filter)
+          : fieldIdToFilters[filter.fieldId] = [filter];
+      }
+      return fieldIdToFilters;
+    }, {}
+  );
+
+  const groupedFilterLabels = Object.keys(fieldIdToFilters);
+  if (groupedFilterLabels.length === 1 && combinationFilters.length === 0) {
```
yen-tt, line 60: how about the case that we only have 1 item in combinationFilters, we should just return that without making a conjunction right?

tests/unit/search-headless.ts
```diff
@@ -1,8 +1,8 @@
-import { Matcher, QuerySource, QueryTrigger } from '@yext/search-core';
+import { DisjunctionStaticFilter, FieldValueStaticFilter, FilterCombinator, Matcher, QuerySource, QueryTrigger } from '@yext/search-core';
```
yen-tt, line 1: unused imports

src/utils/filter-utils.ts
nmanu1: I originally thought that it might be better to differentiate based on order because people may want to display the combined filters based on that. But, thinking about it more, I don't think Headless should be dictated by display concerns, and these filter arrays still technically result in the same filter. I'll update to remove the checks on the order!

src/utils/transform-filters.ts
nmanu1: this case is handled by line 41 where there is only 1 selected filter. the reason a special check is needed for the grouped filters is because those could have started as multiple selected filters, but then been grouped into a single filter combined with OR

tests/unit/slices/filters.ts
oshi97: We can keep them! Was wondering what we prefer as a team

src/utils/transform-filters.ts
```diff
@@ -1,60 +1,70 @@
...
+          ? fieldIdToFilters[filter.fieldId].push(filter)
+          : fieldIdToFilters[filter.fieldId] = [filter];
+      }
+      return fieldIdToFilters;
+    }, {}
+  );
+
+  const fieldIds = Object.keys(fieldIdToFilters);
+  if (fieldIds.length === 1 && combinationFilters.length === 0) {
+    return combineFiltersWithOR(fieldIdToFilters[fieldIds[0]]);
```
oshi97, line 61: If we use Object.values instead of `const fieldIds = Object.keys(fieldIdToFilters);` this can be `combineFiltersWithOR(filters)`

oshi97: we also won't have to call Object.values an additional time

nmanu1: updated!

oshi97 (approved):

PR #164, opened 2024-01-23
Remove real live api key

EmilyZhang777:
This change removes the real live api key to resolve a vulnerability.

J=VULN-37164
TEST=auto

### DIFF
--- a/test-site-node/index.js
+++ b/test-site-node/index.js
@@ -1,7 +1,7 @@
 const { provideHeadless } = require('@yext/search-headless');
 
 const answers = provideHeadless({
-  apiKey: '2d8c550071a64ea23e263118a2b0680b',
+  apiKey: '<REPLACE ME>',
   experienceKey: 'slanswers',
   locale: 'en'
 });
--- a/test-site-node/index.js
+++ b/test-site-node/index.js
@@ -1,7 +1,7 @@
 const { provideHeadless } = require('@yext/search-headless');
 
 const answers = provideHeadless({
-  apiKey: '<REPLACE ME>',
+  apiKey: process.env.API_KEY,
   experienceKey: 'slanswers',
   locale: 'en'
 });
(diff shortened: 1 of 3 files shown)


test-site-node/index.js
```diff
@@ -1,7 +1,7 @@
 const { provideHeadless } = require('@yext/search-headless');
 
 const answers = provideHeadless({
-  apiKey: '2d8c550071a64ea23e263118a2b0680b',
```
benmcginnis, line 4: can we modify this to use an environment variable, e.g. `process.env.API_KEY` that should allow us to do `API_KEY=realkey npm run start`

EmilyZhang777: Updated!

nmanu1 (commented):
technically, removing the api key like this won't really help since the repo (and it's past versions) are open source. if security is concerned about these keys getting out, it might be worth looking into whether github has a way to fully redact/delete part of a file, including its past versions

EmilyZhang777:
I'm planning on deactivating the api keys, so this change should be sufficient.

nmanu1 (commented):
> I'm planning on deactivating the api keys, so this change should be sufficient.

would that deactivate the slapshot test account? I find that pretty useful for testing and I believe a lot of our SDKs' test sites are built around the configuration of that experience. it would also be worth checking if the public-facing search-ui-react Storybook site is based off of this experience as well

EmilyZhang777:
We could generate a new api key for the account/experience and use that in places needed without deactivate the account/experience itself.

nmanu1 (approved):

nmanu1 (commented):
if this is a hotfix version, you'll want to merge into a hotfix branch rather than develop

EmilyZhang777:
I'll update search-core first and revisit this PR to include version bump for search-core 👍

EmilyZhang777:
Yeah, talked to mcgin and confirmed we still want to remove them

PR #167, opened 2024-01-30
Bump search-core to v2.5.51

EmilyZhang777:


nmanu1 (commented):
do we need to make a hotfix for search-headless? none of the vulnerabilities fixed in search-core affect what's actually exported by the package. they were all in the test-site or for devDependencies. also, new users of search-headless would automatically get the latest patch of search-core without this update since we don't pin to a specific patch

EmilyZhang777:
That's fair. I'll abandon this PR then.

PR #174, opened 2024-06-28
Reuse queryId for pagination

Fondryext:
This change adjusts our vertical querying logic to re-use the same queryId when the offset is non-zero or the previous query matches the current input, as both are indications that pagination is being used.
J=WAT-4105
TEST=auto

wrote new tests

### DIFF
--- a/src/search-headless.ts
+++ b/src/search-headless.ts
@@ -411,6 +411,9 @@ export default class SearchHeadless {
     const { limit, offset, sortBys, locationRadius } = this.state.vertical;
     const { referrerPageUrl, context } = this.state.meta;
     const { userLocation } = this.state.location;
+    const zeroOffset = this.state.vertical.offset ?? 0;
+    const inputIsPreviousQuery = input == this.state.query.mostRecentSearch;
+    const queryId = !zeroOffset || inputIsPreviousQuery ? this.state.query.queryId : undefined;
 
     const facetsToApply = facets?.map(facet => {
       return {
@@ -437,7 +440,8 @@ export default class SearchHeadless {
       context,
       referrerPageUrl,
       locationRadius,
-      additionalHttpHeaders: this.additionalHttpHeaders
+      additionalHttpHeaders: this.additionalHttpHeaders,
+      queryId
     };
 
     let response: VerticalSearchResponse;
--- a/src/search-headless.ts
+++ b/src/search-headless.ts
@@ -402,7 +402,7 @@ export default class SearchHeadless {
       return;
     }
     this.stateManager.dispatchEvent('searchStatus/setIsLoading', true);
-    const { input, querySource, queryTrigger } = this.state.query;
+    const { input, querySource, queryTrigger, mostRecentSearch, queryId } = this.state.query;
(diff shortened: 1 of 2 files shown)


src/search-headless.ts
```diff
@@ -411,6 +411,9 @@ export default class SearchHeadless {
     const { limit, offset, sortBys, locationRadius } = this.state.vertical;
     const { referrerPageUrl, context } = this.state.meta;
     const { userLocation } = this.state.location;
+    const zeroOffset = this.state.vertical.offset ?? 0;
+    const inputIsPreviousQuery = input == this.state.query.mostRecentSearch;
```
EmilyZhang777, line 415: nit, can you access `mostRecentSearch` and `queryId` on line 409?

src/search-headless.ts
```diff
@@ -411,6 +411,9 @@ export default class SearchHeadless {
...
+    const queryId = !zeroOffset || inputIsPreviousQuery ? this.state.query.queryId : undefined;
```
EmilyZhang777, line 416: `(!zeroOffset || inputIsPreviousQuery) ? queryId : undefined` to make it clearer?

EmilyZhang777 (commented):
Generally looks good! Just a few nits.

src/search-headless.ts
```diff
@@ -437,7 +440,8 @@ export default class SearchHeadless {
       context,
       referrerPageUrl,
       locationRadius,
-      additionalHttpHeaders: this.additionalHttpHeaders
+      additionalHttpHeaders: this.additionalHttpHeaders,
+      queryId
```
vijay267, line 444: nit: Similar to Emily's comment above, if possible I'd prefer if we could have additionalHttpHeaders as the last parameter, and thus should add queryId before it. Basically have all the very specific params first and then the additional http headers.

tests/integration/verticalsearch.ts
```diff
@@ -169,6 +169,24 @@ it('answers.setVerticalLimit sets the vertical limit when a number is passed to
   expect(answers.state.vertical.limit).toEqual(7);
 });
 
+it('vertical searches re-use queryId when offset is non-0', async () => {
```
vijay267, line 172: Maybe this is already covered by another test, but should we have a test where no query id is passed? Like offset = 0 and input query is different?

vijay267 (commented):
Some minor comments but generally looks good

src/search-headless.ts
Fondryext: Done!

Fondryext: yup, adjusted

src/search-headless.ts
Fondryext: sure thing

src/search-headless.ts
Fondryext: done

tests/integration/verticalsearch.ts
Fondryext: Sure, added. Also fixed the tests overall (queryId wasn't getting generated)

Fondryext:
Decided to not go with this approach, so closing.

PR #178, opened 2024-08-21
search-headless: increase search-core version used for function vertical support

alapre:
Increases search-core version to 2.6.0-beta.3 (and increases search-headless beta version too)

Will also create PRs for search-headless-react and search-ui-react to support function vertical.

Tests: auto, manual

Edited index.js in test-site-node (reverted this before pushing) to confirm function vertical was working as expected. Also ran `npm run test` to ensure no tests failed.


Fondryext (approved):
It's too bad about all the .md files, but since this is a very small PR otherwise I'm fine to just pull them in here.

PR #180, opened 2024-09-12
Add CloudChoice Support

k-gerner:
Adds support for cloudChoice.

J=WAT-4377
TEST=compile

Ran `npm run build`


THIRD-PARTY-NOTICES
```diff
@@ -62,7 +62,7 @@ SOFTWARE.
 
 The following NPM package may be included in this product:
 
- - @yext/search-core@2.5.0
+ - @yext/search-core@2.5.4
```
vijay267, line 65: I'm pretty sure we want to merge this into a hotfix branch first and not directly into main / master right?

src/search-core-re-exports.ts
```diff
@@ -16,6 +16,7 @@ export {
   Context,
   ClientSDKHeaderValues,
   CloudRegion,
+  CloudChoice,
```
vijay267, line 19: Since this is mostly alphabetical (it's not entirely though), should choice be before region then?

THIRD-PARTY-NOTICES
k-gerner: Huh, I assumed it would automatically create the hotfix branch since I have that as upstream on my local. Updated

src/search-core-re-exports.ts
k-gerner: Done

vijay267 (approved):

PR #188, opened 2024-11-22
search-headless: Resolve Vulnerabilities

mkouzel-yext:
Regenerated package-lock.json and test-site-node/package-lock.json to ensure they include updated safe dependencies.

J=VULN-39420
TEST=manual

Looked at babel dependencies in both package-lock files and made sure they had the safe versions.


Fondryext (approved):
Pretty sure this is all good

PR #197, opened 2025-04-01
Version 2.6.1

Fondryext:
### Changes

- Adds experienceKey and locale to the MetaState, allowing them to be referenced later #196


src/models/slices/meta.ts
```diff
@@ -24,5 +24,13 @@ export interface MetaState {
   /**
    * Indicates the type of search that Search Headless is managing.
   */
-  searchType: SearchType
-}
\ No newline at end of file
+  searchType: SearchType,
+  /**
+   * Indicates the name of the experience that Search Headless is managing.
```
k-gerner, line 29: nit: name technically refers to something else, can we change this to key / id or similar

src/search-headless.ts
```diff
@@ -52,11 +53,15 @@ export default class SearchHeadless {
...
+    private config: HeadlessConfig,
     private core: SearchCore,
     private stateManager: StateManager,
     private httpManager: HttpManager,
     private additionalHttpHeaders?: AdditionalHttpHeaders
-  ) {}
+  ) {
+    this.stateManager.dispatchEvent('meta/setExperienceKey', config.experienceKey);
+    this.stateManager.dispatchEvent('meta/setLocale', config.locale);
+  }
```
k-gerner, line 64: Just for my own understanding, are the other properties in the meta state (e.g. search type) set at query time?

src/models/slices/meta.ts
Fondryext: Sure thing. Also added disclaimers.

k-gerner (dismissed):

k-gerner (approved):

PR #205, opened 2025-07-08
Add name property for near field value filters

mkouzel-yext:
Bumps search-core version to include change that added optional name property for near field value filters.


k-gerner (commented):
Maybe we should bump the `search-core` version explicitly?

k-gerner (approved):

anguyen-yext2 (approved):

